PR #18, opened 2018-01-17
set AWS_PROFILE to arg

nzoschke:
I'm building a utility invoked by `aws-okta`. It would be useful for the utility to know what profile its running under.

Current behavior:

```
$ aws-okta exec prod -- env | grep prod
```

New behavior:

```
$ aws-okta exec prod -- env | grep prod
AWS_PROFILE=prod
```


ejcx:
Just going to merge this since you both have probably forgot about it =]

PR #21, opened 2018-02-24
reuse session cookie:

dfuentes:
- Grab session cookie and store in keyring.  Attempt to reuse that
  session cookie when requesting saml assertions

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -36,6 +36,8 @@ type OktaClient struct {
 	SessionToken    string
 	Expiration      time.Time
 	OktaAwsSAMLUrl  string
+	CookieJar       *cookiejar.Jar
+	BaseURL         *url.URL
 }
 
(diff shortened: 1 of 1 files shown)


lib/okta.go
```diff
@@ -36,6 +36,8 @@ type OktaClient struct {
 	SessionToken    string
 	Expiration      time.Time
 	OktaAwsSAMLUrl  string
+	CookieJar       *cookiejar.Jar
```
achille-roussel: How about using the `http.CookieJar` interface here instead of the plain type to make the code more flexible? https://golang.org/pkg/net/http/#CookieJar

ejcx:
Really solid improvement! I like it a lot.

lib/okta.go
dfuentes: sure 2450b3a

PR #23, opened 2018-03-01
Less auth prompts

dfuentes:
- Use fork of keyring that allows setting ACL's correctly
- Use login keychain


abraithwaite:
giphy 13

PR #27, opened 2018-03-19
Allow for multiple Okta Apps

elementalvoid:
Get the aws_saml_url from profiles -- with fallback to "okta" section.

The `aws_saml_url` can now be set in the "okta" ini section, or on a per profile basis. This is useful if, for example, your organization has several Okta Apps (i.e. one for dev/qa and one for prod, or one for internal use and one for integrations with third party providers).


dfuentes:
Thanks, this looks great.  Extra thanks for documenting it :)

Issue #17, opened 2018-01-11
login: Allow user to specify a browser path


otterley:
Sometimes I like to use multiple browsers for logging into AWS (for example, when I want to compare two different environments, like prod and stage).  

It would be useful to be able to specify that I'd like to log in via Safari (which is not my default browser) instead of Chrome for a particular session.  I'm an OS X user.

ejcx:
@otterley `aws-okta` is just using https://github.com/skratchdot/open-golang/, which is wrapper around `open`.

This would be pretty major surgery on the `login` subcommand.

dfuentes:
The workaround [here](https://github.com/segmentio/aws-okta/pull/29) has been merged, that may work for your use case

otterley:
Good enough, thanks!

PR #29, opened 2018-03-20
allow login url to go to stdout

nicgrayson:
Workaround for #17 

This is just like [aws-vault](https://github.com/99designs/aws-vault/blob/master/cli/login.go#L187) and allows other browsers to be called via shell configuration like this:

```shell
cat `which awslogin`
#!/bin/bash

aws-okta login $1 --stdout | xargs -t /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --incognito --new-window
```

This is borrowed from: https://github.com/FernandoMiguel/kb/blob/master/Chrome/aws-vault.md

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -20,8 +20,12 @@ var loginCmd = &cobra.Command{
 	RunE:  loginRun,
 }
 
+// Stdout is the bool for -stdout
+var Stdout bool
+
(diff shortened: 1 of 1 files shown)


dfuentes:
Thanks for the contribution

PR #30, opened 2018-03-20
allow role and saml-provider ARNs to appear in either order

hopkinsth:
Long commit message is long. Hi there! Thanks for building this tool. I put the same summary of this issue in a comment in the code, and it describes my intent here...

Amazon's documentation suggests that the Role ARN should appear first in the comma-delimited set in the Role Attribute that SAML IdP returns.

See the section titled "An Attribute element with the Name attribute set to https://aws.amazon.com/SAML/Attributes/Role" on [this page](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_saml_assertions.html). 

In practice, though, Okta SAML integrations with AWS will succeed with either the role or principal ARN first, and the `if` statements I adjusted allow that behavior in this program.

The AWS accounts I'm trying to use with this tool are set up that way: the role ARN comes first, so existing releases of this tool print a message showing that my base role "is not authorized by Okta," which is not the case.

### DIFF
--- a/lib/utils.go
+++ b/lib/utils.go
@@ -21,9 +21,25 @@ func GetRoleFromSAML(resp *saml.Response, profileARN string) (string, string, er
 					continue
 				}
 
-				principalARN, profile := tokens[0], tokens[1]
-				if profile == profileARN {
-					return principalARN, profile, nil
+				// Amazon's documentation suggests that the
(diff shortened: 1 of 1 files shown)


dfuentes:
Thank you for the thorough report, documentation, and contribution!

PR #35, opened 2018-03-30
fix password prompt error in windows

richardheath:
Tried compiling tool for windows and encountered these issues before getting it to work.

First I need to fix keyring dependency: [PR](https://github.com/99designs/keyring/pull/19)

After I am able to compile this line failed with "The handle is not defined" in windows. Used syscal.Stdin as suggested on this [thread](https://github.com/golang/go/issues/11914):

I tested changes on Windows and Mac.

### DIFF
--- a/lib/prompt.go
+++ b/lib/prompt.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"syscall"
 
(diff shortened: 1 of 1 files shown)


dfuentes:
Looks good, thanks for the fix

PR #39, opened 2018-04-10
Add analytics

dfuentes:
Adds some very basic usage analytics.  No analytics are sent by default, and they can only be enabled via linker flags at compile time (which we do not do for public github releases).


cmd/root.go
```diff
@@ -20,24 +21,33 @@ var (
...
+	backend          string
+	debug            bool
+	version          string
+	analyticsEnabled bool
+	analyticsClient  analytics.Client
+	username         string
+)
+
+const (
+	AnalyticsWriteKey = "qK4bR1ogitMozAAFUw3om77inuUjvwbJ"
```
f2prateek: left from debugging? not sure if we want this to be in the open source version

dfuentes: Write keys are public right?  I mean, we put them in javascript...

f2prateek: Yep, they can be - depends on how strict you want to be.

PR #42, opened 2018-04-18
Add support for Okta push notifications

erickt:
This refactors the post challenge to add support for Okta push notifications by only contacting duo when the MFA factor is "DUO". I don't have a setup with Duo support, so I wasn't able to verify if this works with Duo though.


dfuentes:
Thanks for the improvement.  I tested against a DUO push enabled okta account and it works for that as well.

PR #44, opened 2018-05-01
Add Duration flags to login cmd

reverson:
Add duration flags for login cmd, and increase maxAssumeRoleDuration to new max allowed value in AWS


dfuentes:
Looks good, thanks for the contribution

PR #45, opened 2018-05-04
Update readme to match config parsing code

gregburek:
The README regarding config appears to have drifted from the code, as
one must use 'source_profile' to indicate which role one uses as a hub
to assume another. This commit updates the README example with the
proper config key.

Also, 'role_arn' is added to the integrations-auth example profile, as
it is required.


dfuentes:
Thanks for the update

PR #46, opened 2018-05-04
Allows for form based Duo Authentication Fixes #26

### LINKED ISSUE #26: Authentication flow fails at Duo Push, unsure of cause
I've set up an aws-okta profile, and can use Okta to log into the role it specifies normally. When I attempt to use aws-okta, it gets to the MFA portion, waits for 2-3 seconds, then fails without sending a push. From the Duo side, there are no corresponding failed MFA events. Is there anything else I can do track down what's going on?
```
$ aws-okta login profile-name -d
DEBU[0000] Parsing config file /Users/andrewr/.aws/config
DEBU[0000] using okta provider
DEBU[0000] Failed to reuse session token, starting flow from start
DEBU[0000] Step: 1
DEBU[0002] Step: 2
INFO[0002] Sending push notification...
DEBU[0002] Okta Factor ID: <looks right>

DEBU[0003] Host:api-23a196fe.duosecurity.com
Signature:<REDACTED>
StateToken:<REDACTED>

DEBU[0003] challenge u2f
Failed Duo challenge
```

gregburek:
In some integrations, it appears that hitting /frame/web/v1/auth with an
empty POST will not result in a 302, as expected, but returns a 200,
with a response body that contains a form with hidden input fields, sid
and certs_url. In order to continue with authentication, this form data
must be POSTed to the same /frame/web/v1/auth, which then triggers the
already handled 302 redirect flow.

This commit adds a branch to DoAuth where on StatusOK, we parse out the
values of the sid and certs_url inputs. These values are then passed
recursively to DoAuth again, so that another pass through the function
may submit these values and handle the expected 302 redirect flow. This
was found to be the order of operations needed from chrome dev tools
inspection of the iframe API calls, and it is working for our
integration setup.

It was attempted to use the form's sid value instead of trying to get it
from the 302 redirect location, but that resulted in a 403 Forbidden
from DoPrompt.

Also, GetNode has been generalized, in order to reuse the input form
parsing for sid and certs_url in addition to SAMLResponse.

### DIFF
--- a/lib/duo.go
+++ b/lib/duo.go
@@ -6,6 +6,10 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+
+	uniformResourceLocator "net/url"
+
+	"golang.org/x/net/html"
 )
 
 type DuoClient struct {
@@ -64,7 +68,7 @@ func (d *DuoClient) ChallengeU2f() (err error) {
 
 	tx = strings.Split(d.Signature, ":")[0]
 
-	sid, err = d.DoAuth(tx)
+	sid, err = d.DoAuth(tx, "", "")
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


andrewrosezen:
This works, FWIW. I suspect that the differing behavior is due to Duo Beyond's Trusted Endpoint configuration. I was able to get my policy to work by statically setting my User agent to the same string that Chrome sends (mega hacky and fragile, do not recommend).

dfuentes:
Ah, the fun of reverse engineering the iframe code.  Thanks for this contribution, the code looks good to me (and I have verified that it continues to work with our Duo integration as well)

PR #54, opened 2018-05-21
Fix issuer name

eperdeme:
Changed vault -> Okta. Minor thing but confused me slightly when I got this page after timeout in AWS;

"Amazon Web Services Sign In
Your session has expired.To start a new session, click the link provided by your administrator:aws-vault To logout, click here"


dfuentes:
good catch

PR #56, opened 2018-06-28
Validate okta creds on `aws-okta add`

nickatsegment:
- split user auth out of `AuthenticateProfile` into `AuthenticateUser`
- use `AuthenticateUser` to validate creds

See TOOL-340

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -68,6 +68,11 @@ func add(cmd *cobra.Command, args []string) error {
 		Password:     password,
 	}
 
+	if err := creds.Validate(); err != nil {
+		log.Debugf("Failed to validate credentials: %s", err)
+		return ErrFailedToValidateCredentials
+	}
+
 	encoded, err := json.Marshal(creds)
 	if err != nil {
 		return err
@@ -81,6 +86,7 @@ func add(cmd *cobra.Command, args []string) error {
 	}
 
 	if err := kr.Set(item); err != nil {
+		log.Debugf("Failed to add user to keyring: %s", err)
 		return ErrFailedToSetCredentials
 	}
 
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -53,6 +53,20 @@ type OktaCreds struct {
 	Password     string
 }
 
+func (c *OktaCreds) Validate() error {
+	// OktaClient assumes we're doing some AWS SAML calls, but Validate doesn't
+	o, err := NewOktaClient(*c, "", "")
+	if err != nil {
+		return err
+	}
+
+	if err := o.AuthenticateUser(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func NewOktaClient(creds OktaCreds, oktaAwsSAMLUrl string, sessionCookie string) (*OktaClient, error) {
 	base, err := url.Parse(fmt.Sprintf(
 		"https://%s.%s", creds.Organization, OktaServer,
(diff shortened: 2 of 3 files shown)


lib/okta.go
```diff
@@ -85,44 +100,57 @@ func NewOktaClient(creds OktaCreds, oktaAwsSAMLUrl string, sessionCookie string)
...
-		log.Debug("Step: 1")
-		err = o.Get("POST", "api/v1/authn", payload, &oktaUserAuthn, "json")
-		if err != nil {
-			return sts.Credentials{}, "", errors.New("Failed to authenticate with okta.  Please check that your credentials have been set correctly with `aws-okta add`")
+	o.UserAuth = &oktaUserAuthn
+
+	// Step 2 : Challenge MFA if needed
+	log.Debug("Step: 2")
+	if o.UserAuth.Status == "MFA_REQUIRED" {
+		log.Info("Requesting MFA")
```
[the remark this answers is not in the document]
nickatsegment: ✅

nickatsegment:
@dfuentes: we talked IRL about saving/reusing the session cookie. Seems like there's a few cookies/tokens in play here. You get the token from the Okta auth call, then you use the token in the AWS SAML call, and that gives you a reusable cookie. Detangling the AWS SAML stuff so it's usable from `add` would be a fair amount of work at this point. It'd probably be easier to roll `add` into the `login` command. I'm gonna leave it as future work

cmd/add.go
```diff
@@ -68,6 +68,10 @@ func add(cmd *cobra.Command, args []string) error {
 		Password:     password,
 	}
 
+	if err := creds.Validate(); err != nil {
+		return fmt.Errorf("Failed to validate credentials: %s", err)
```
systemizer: any reason not to use errfailedtosetcredentials error here?

lib/okta.go
```diff
@@ -85,44 +99,55 @@ func NewOktaClient(creds OktaCreds, oktaAwsSAMLUrl string, sessionCookie string)
...
+	if err != nil {
+		log.Debug("Failed to reuse session token, starting flow from start")
 
-		if o.UserAuth.SessionToken == "" {
-			return sts.Credentials{}, "", fmt.Errorf("authentication failed for %s", o.Username)
+		if err := o.AuthenticateUser(); err != nil {
+			return sts.Credentials{}, "", err
 		}
 
 		// Step 3 : Get SAML Assertion and retrieve IAM Roles
```
systemizer: let's update and not specify step 3 since that doesn't make sense in context anymore

cmd/add.go
nickatsegment: Yeah, I wanted to include the upstream error.

nickatsegment: I guess that means the ErrFailedToSetCredentials definition is redundant now, so I'll take that out

nickatsegment:
Re https://github.com/segmentio/aws-okta/pull/56#pullrequestreview-133721623 I experimented with cleaning it up a bit. I'll add the commit and you can tell me what you think

cmd/add.go
nickatsegment: https://github.com/segmentio/aws-okta/pull/56/commits/997fc6691b63ecf8b93529eec7743048d87f0a72

systemizer: oh! i meant adding additional errors. It seems the pattern in this repo is to reference errors instead of creating errors inline. I'm guessing this is useful for others who might want to condition their code on what type of error is returned. Could we add an error like ErrInvalidCredentials and return if validate fails? and for the one below, return the original "ErrFailedToSetCredentials"? sorry for the confusion

systemizer:
nice work!

cmd/add.go
```diff
@@ -68,6 +68,11 @@ func add(cmd *cobra.Command, args []string) error {
 		Password:     password,
 	}
 
+	if err := creds.Validate(); err != nil {
+		log.Debugf("Failed to validate credentials: %s", err)
```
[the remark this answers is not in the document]
boggsboggs: I don't think there are a lot of programming rules that hold for all possible applications. When developing an HTTP server, CLI, language parser, library, REPL playground, etc, you'll often want different designs or conventions.

My experience is that language developers tend to specify design principles appropriate for developing libraries (which is what they do). They're useful to know so you can have accurate expectations about library behavior, and you should follow them when writing code that is likely to be called by other developers ("library" code).

When you're building an application that is't a library, everything depends on what job the application is trying to accomplish. Being usable as a library is usually a very low priority compared to getting the job done with minimal effort.

In this particular case, I'd say that we want to log the error and return it. Logging the error will make it easier for us to debug. No one calls this function without reading its body first ("not a library"); we have only the aws-okta developers to consider. And, we need to return the error so calling code knows to fail.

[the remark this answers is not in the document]
nickatsegment: 100% agreed. I wasn't really considering making changes, it's just a funny coincidence.

PR #57, opened 2018-06-29
Add support for assume_role_ttl in config profiles

devonbleak:
As AWS allows for configuration of maximum session duration per-role it seems appropriate and useful to be able to set this per-profile in the AWS config.

Also adds documentation for the environment variables `AWS_SESSION_TTL` and `AWS_ASSUME_ROLE_TTL`.


nickatsegment:
@devonbleak 	sorry for the delay. Looks good to me. Thanks!

PR #59, opened 2018-06-29
Add windows support docs

devonbleak:


snag-aravind-kandalam:
is this going to be merged?

nickatsegment:
I don't have a Windows box to text, but I wonder if the deb packages would work in WSL? https://github.com/segmentio/aws-okta/wiki/Installation

devonbleak:
It does generally work in WSL as expected however WSL doesn't really work with the rest of Windows and the filesystems as expected (home directories are different, no GUI apps, etc) so at this point we generally recommend that our users use a native Windows build as it tends to be less confusing, but lately getting those built and the dependencies installed is pretty painful.

I'll try to get another PR submitted with updates from our internal docs.

PR #60, opened 2018-07-07
Add Okta server selection

### LINKED ISSUE #48: Can we use okta orgs in the okta-emea region
Hi,

Is it possible to use Okta organizations in the emea region? I don't see a way to provide this when executing the aws-okta add command?

Thanks,

Jeffrey

neuralsandwich:
Add support for selecting Okta region, closes #48.

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -46,6 +46,11 @@ func add(cmd *cobra.Command, args []string) error {
 	}
 
 	// Ask username password from prompt
+	server, err := lib.Prompt("Okta Region (emea/us)", false)
+	if err != nil {
+		return err
+	}
+
 	organization, err := lib.Prompt("Okta organization", false)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


peter-murray:
You are still excluding the Okta preview accounts with this change... if you are serious about making any change to support this properly you have to consider the preview accounts, and accounts that can operate under a custom URL, which would be a URL not using any of the three known Okta domains.

NeuralSandwich:
@peter-murray Thanks for the feedback. I haven't really interacted too much with Okta, I didn't know about these other options. I'm happy to review this change and take these into consideration.

yverry:
Works well as expected.
As peter said, this is not the ultimate solution but ... the beginning to solve the okta domains problem.

From my side I prefer to merge this PR instead of "wait the golden solution"

NeuralSandwich:
Yeah, totally agree the prompt is pretty crap. I had literally just picked Go up when I wrote this and did the first thing that came to mind. I wasn't really wanting to mess with this too much. I'll try to write up something with a bit of a cleaner solution. I'm using this modification currently, it _works_ but is less than optimal.

nickatsegment:
Pretty sure this is superseded by #92.

PR #61, opened 2018-07-11
Add an AWS_OKTA_PROFILE env attribute

logikal:
Useful for when you have tooling that wants to know which role you're running in.

We don't want to use AWS_PROFILE, since that's significant to some tools.
Instead, add an AWS_OKTA_PROFILE env var, which is the name of the profile being used.


dfuentes:
lgtm, thanks for the addition

logikal:
This was approved a bit ago, is there anything I need to do to have it merged?

logikal:
@dfuentes is there anything I can do to get this merged into the released binary?

dfuentes:
Sorry for the delay

logikal:
Thanks, @dfuentes !

PR #66, opened 2018-08-03
MFA input validation

Incisive:
Addressing some possible confusion around current “index out of range” runtime errors when selecting an MFA option.


nickatsegment:
Thanks @Incisive! I updated this and shipped it in #139. Sorry for the delay.

PR #67, opened 2018-08-13
add a `list` command

logikal:
I can never remember the names of the profiles I have configured, so this command will list them for you.

Caveat: It assumes you only care about profiles that are assuming from another role; it only lists profiles that have a source_profile.

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"sort"
 	"text/tabwriter"
 
 	analytics "github.com/segmentio/analytics-go"
@@ -32,11 +33,20 @@ func listRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Let's sort this list of profiles so we can have some more deterministic output:
+	var profileNames []string
+
+	for profile := range profiles {
+		profileNames = append(profileNames, profile)
+	}
+
+	sort.Strings(profileNames)
+
 	w := new(tabwriter.Writer)
 	w.Init(os.Stdout, 0, 8, 2, '\t', 0)
-	fmt.Fprintln(w, "profile\tarn\tsource_role\t")
-	fmt.Fprintln(w, "---\t---\t---\t")
-	for profile, v := range profiles {
+	fmt.Fprintln(w, "PROFILE\tARN\tSOURCE_ROLE\t")
+	for _, profile := range profileNames {
+		v := profiles[profile]
 		if role, exist := v["role_arn"]; exist {
 			fmt.Fprintf(w, "%s\t%s\t%s\n", profile, role, v["source_profile"])
 		}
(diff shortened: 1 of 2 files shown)


logikal:
@dfuentes Is this a feature ya'll would like?

nickatsegment:
@logikal 	Yeah, I think it could be useful. Let me take a look.

cmd/list.go
```diff
@@ -0,0 +1,64 @@
...
+	Use:   "list",
+	Short: "list will show you the profiles currently configured",
+	RunE:  listRun,
+}
+
+func init() {
+	RootCmd.AddCommand(listCmd)
+}
+
+func listPre(cmd *cobra.Command, args []string) {
```
nickatsegment: Dead code?

cmd/list.go
```diff
@@ -0,0 +1,64 @@
...
+	w.Flush()
+
+	if analyticsEnabled && analyticsClient != nil {
+		analyticsClient.Enqueue(analytics.Track{
+			UserId: username,
+			Event:  "Listed Profiles",
+			Properties: analytics.NewProperties().
+				Set("backend", backend).
+				Set("aws-okta-version", version).
+				Set("profile_count", len(profiles)).
```
nickatsegment: I think this should be `profile-count`, to standardize on dashes being the word separator. @systemizer 	may have opinions.

logikal:
Thanks, @nickatsegment! I've fixed up the feedback and updated from master, squashed it all up, and think it's ready to go now. 🎅

logikal:
Anything else needed here, @nickatsegment ?

nickatsegment:
Sorry for the delay!

cmd/list.go
```diff
@@ -0,0 +1,61 @@
...
+
+	w := new(tabwriter.Writer)
+	w.Init(os.Stdout, 0, 8, 2, '\t', 0)
+	fmt.Fprintln(w, "profile\tarn\tsource_role\t")
+	fmt.Fprintln(w, "---\t---\t---\t")
+	for profile, v := range profiles {
+		if role, exist := v["role_arn"]; exist {
+			if src, exist := v["source_profile"]; exist {
+				s := fmt.Sprintf("%s\t%s\t%s\t", profile, role, src)
+				fmt.Fprintln(w, s)
```
nickatsegment: Just use `fmt.Fprintf` directly?

cmd/list.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+
+	w := new(tabwriter.Writer)
+	w.Init(os.Stdout, 0, 8, 2, '\t', 0)
+	fmt.Fprintln(w, "profile\tarn\tsource_role\t")
+	fmt.Fprintln(w, "---\t---\t---\t")
+	for profile, v := range profiles {
+		if role, exist := v["role_arn"]; exist {
+			if src, exist := v["source_profile"]; exist {
```
nickatsegment: In the case where there is no `source_profile`, no line gets printed. This is the case in my config, so I get no results from the whole command :( I wonder if you could use `lib/config.go`'s `sourceProfile` function?

logikal:
Thanks, @nickatsegment! 

I fixed this to display profiles that don't have a source_profile, and also have it sort by `profile_name` so it gives the same order on each run.

nickatsegment:
🚢

PR #73, opened 2018-08-30
Implement config option for which MFA to use

dnelson:
When users have multiple MFA types configured, they are prompted each time for which one they want to use. This introduces unnecessary friction for the normal case, especially when using the Okta app, which is both a push-based MFA and also TOTP, and will thus force the prompt even when it is the only configured MFA type.

This PR adds two additional config options to the `[okta]` config block: `mfa_factor` which is the MFA provider, and `mfa_type`, which is the kind of mfa it is. For example, the Okta app push notification is defined as follows:
```
[okta]
aws_saml_url = home/amazon_aws/0ac4qfegf372HSvKF6a3/965
mfa_factor = OKTA
mfa_type = push
```


dnelson:
@nickatsegment anything I can do to unstick this PR? I’d like to be able to switch our org back to the upstream release

lsowen:
@dnelson I was just looking into implementing something similar, as I have multiple MFA methods and would like it to not prompt me every time.  

However, one thing I noticed about this PR is that the config is based on "factor type" and "factor provider".  Is it possible to have multiple entries for the same "type"/"provider" pair?  (I know it is at least possible to have multiple of the same "type" for different "providers").  If so, maybe this could be based on [factor `Id`](https://github.com/segmentio/aws-okta/blob/master/lib/struct.go#L29) instead? 

A side benefit is that it would be easy to add an environment variable like `OKTA_MFA_ID` to optionally override the config setting.

nickatsegment:
Thanks for the PR @dnelson! It'd take a fair amount of work for me to test this since we don't really use this feature at Segment AFAIK. Not that I won't do it, it just might take a while. I'd like to hear from other users that have this use case. (https://github.com/segmentio/aws-okta/pull/73#issuecomment-422922232 is a start.)

Also, I wonder how this interacts with https://github.com/segmentio/aws-okta/pull/83?

deiwin:
> Also, I wonder how this interacts with #83?

Should be easy to merge. As MFA is handled by the browser for OIDC apps, there's nothing to do about MFA in code for the new client. The old client can be modified as is done here.

lib/okta.go
```diff
@@ -67,7 +68,7 @@ func (c *OktaCreds) Validate() error {
 	return nil
 }
 
-func NewOktaClient(creds OktaCreds, oktaAwsSAMLUrl string, sessionCookie string) (*OktaClient, error) {
+func NewOktaClient(creds OktaCreds, oktaAwsSAMLUrl string, sessionCookie string, mfaConfig MFAConfig) (*OktaClient, error) {
```
nickatsegment: This breaks the API. We're still v0, so I guess that's fine.

lib/provider.go
```diff
@@ -162,6 +162,10 @@ func (p *Provider) getSamlSessionCreds() (sts.Credentials, error) {
 		ProfileARN:      profileARN,
 		SessionDuration: p.SessionDuration,
 		OktaAwsSAMLUrl:  oktaAwsSAMLUrl,
+		MFAConfig: MFAConfig{
+			MFAFactor: p.profiles["okta"]["mfa_factor"],
```
nickatsegment: Can you think of any way to validate these? I put in the wrong thing and I just get 
```
DEBU[0006] Failed to validate credentials: Failed to authenticate with okta: &errors.errorString{s:"POST https://segment.oktapreview.com/api/v1/authn: 401 Unauthorized"}
```

lib/okta.go
lsowen: Would it be better to not pass it to `NewOktaClient` and just set it with `client.MFAConfig = config` when/where you need it set?  (eg in `OktaProvider.Retrieve()`)

nickatsegment: Sure. Or a new constructor like `NewOktaClientWithMFAConfig`

README.md
```diff
@@ -57,6 +57,15 @@ Global Flags:
...
 
+You can optionally specify which MFA type to use, which is useful when multiple are configured, for example:
+```ini
+[okta]
+aws_saml_url = home/amazon_aws/0ac4qfegf372HSvKF6a3/965
+mfa_factor = OKTA
+mfa_type = push
+
+```
+
```
nickatsegment: Document other types? I see 
* `GOOGLE (token:software:totp)`
* `DUO (web)`
here

lib/okta.go
nickatsegment: Regarding your edit, that would be fine. I think asking the caller to add it after the fact isn't the worst.

Re `OKTA_MFA_ID`: I'd rather not have the lib pulling from env vars, seems leaky to me. But in cmd, sure.

lib/okta.go
```diff
@@ -441,6 +452,12 @@ type OktaProvider struct {
 	ProfileARN      string
 	SessionDuration time.Duration
 	OktaAwsSAMLUrl  string
+	MFAConfig       MFAConfig
+}
+
+type MFAConfig struct {
```
nickatsegment: I know nothing else is really documented, but doc strings would be nice (examples!)

PR #77, opened 2018-09-05
fix for duo empty result_url

eddieramirez:
if result_url is empty use existing cookie (if not empty)

### DIFF
--- a/lib/duo.go
+++ b/lib/duo.go
@@ -242,7 +242,11 @@ func (d *DuoClient) DoStatus(txid, sid string) (auth string, err error) {
 	err = json.NewDecoder(res.Body).Decode(&status)
 
 	if status.Response.Result == "SUCCESS" {
-		auth, err = d.DoRedirect(status.Response.ResultURL, sid)
+		if status.Response.ResultURL != "" {
+			auth, err = d.DoRedirect(status.Response.ResultURL, sid)
+		} else if status.Response.Cookie != ""  {
+			auth = status.Response.Cookie
+		}
 	}
 	return
 }
--- a/lib/duo.go
+++ b/lib/duo.go
@@ -244,7 +244,7 @@ func (d *DuoClient) DoStatus(txid, sid string) (auth string, err error) {
 	if status.Response.Result == "SUCCESS" {
 		if status.Response.ResultURL != "" {
 			auth, err = d.DoRedirect(status.Response.ResultURL, sid)
-		} else if status.Response.Cookie != ""  {
+		} else {
 			auth = status.Response.Cookie
 		}
 	}


eddieramirez:
Can I get a review on this? Thanks!

lib/duo.go
```diff
@@ -242,7 +242,11 @@ func (d *DuoClient) DoStatus(txid, sid string) (auth string, err error) {
 	err = json.NewDecoder(res.Body).Decode(&status)
 
 	if status.Response.Result == "SUCCESS" {
-		auth, err = d.DoRedirect(status.Response.ResultURL, sid)
+		if status.Response.ResultURL != "" {
+			auth, err = d.DoRedirect(status.Response.ResultURL, sid)
+		} else if status.Response.Cookie != ""  {
+			auth = status.Response.Cookie
+		}
```
nickatsegment: What about the case where `status.Response.ResultURL == "" && status.Response.Cookie == ""`?

systemizer:
hey @eddieramirez thanks for the PR. 

Since these are undocumented APIs, it's hard to tell whether this case is valid. Do you have a way of reproducing the situation when `result_url` is empty but the response status is successful?

eddieramirez:
@systemizer That's actually the case I'm encountering. I'm a new user to Okta and aws-okta, so I was successfully getting prompted for Duo for but aws-okta was failing nonetheless. 

```
INFO[0001] Sending Push Notification...                 
Failed Duo challenge
```

It turns out I was getting a cookie on the first try with no result_url which jibes with the original flow that was changed by https://github.com/segmentio/aws-okta/pull/68.

With my code changes, I now get..
```
INFO[0001] Sending Push Notification...                 
Completed 256.0 KiB/6.7 GiB (95.9 KiB/s) with 1 file(s) remaining
```

systemizer:
huh interesting, I wonder why we are getting different prompts. Can't reproduce here, but I believe you :) . I think reverting to the previous behavior makes sense if the result_url is not actionable. this lgtm

lib/duo.go
systemizer: this is how it was done previously: https://github.com/segmentio/aws-okta/pull/68/files#diff-2461124d964c26f736e12a1edae3f065L243 . So im okay with this

systemizer: you're right. `else` makes more sense then `else if` here

nickatsegment: I'm going to approve since this is clearly better than before, but this case needs covering IMO

systemizer: @eddieramirez can you make this an `else` instead? once we have that I will merge

eddieramirez: done

nickatsegment: Thanks. Sorry to be a PITA

PR #78, opened 2018-09-06
Document: Don't append .okta.com to org name

elliotmoore:
If you have used other okta aws cli tools such as https://github.com/oktadeveloper/okta-aws-cli-assume-role - you might assume that you need to append .okta.com" to your organization name.

```
# incorrect usage
$ aws-okta add
Okta organization: mycompany.okta.com
Okta username: elliotmoore
Okta password:
INFO[0014] Added credentials for user elliotmoore
```
Results in: 
```
$ aws-okta login blah
Failed to authenticate with okta.  Please check that your credentials have been set correctly with `aws-okta add`
```

making clear that organization name should not be appended with .okta.com
as its already added here: https://github.com/segmentio/aws-okta/blob/master/lib/okta.go#L72

```
# correct usage
Okta organization: mycompany
```
Perhaps a better PR would warn the user during `aws-okta add` or ignore if exists.
For now this open PR might save somebody some time :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ $ go get github.com/segmentio/aws-okta
 $ aws-okta add
 ```
 
-This will prompt you for your Okta organization, username, and password.  These credentials will then be stored in your keyring for future use.
+This will prompt you for your Okta organization, username, and password.  These credentials will then be stored in your keyring for future use.   
+Note: Don't append ".okta.com" to your organization name, it is automatically added by `aws-okta`.
+
 
 ### Exec
 


nickatsegment:
Thanks! Sorry for the delay.

I'd say best UX is to detect a domain and assume okta.com if not present.

See also https://github.com/segmentio/aws-okta/pull/60 and https://github.com/segmentio/aws-okta/pull/92

elliotmoore:
thanks for following up on this and sharing the other PRs!

PR #83, opened 2018-09-17
Add support for app-level MFA

### LINKED ISSUE #63: Can't authenticate if App-level MFA is turned on
**Summary**

If the profile is connected to an Okta application that has [Application Level MFA](https://support.okta.com/help/Documentation/Knowledge_Article/Application-Level-Multifactor-Authentication-1300941931) enabled, authentication fails.

**OS**: OS X 10.13.6
**aws-okta version**: [aws-okta-v0.19.0-darwin-amd64](https://github.com/segmentio/aws-okta/releases/download/v0.19.0/aws-okta-v0.19.0-darwin-amd64) binary

**Details**

Prerequisites:

- Two Okta applications: one regular, one with MFA turned on.
okta-app-mfa Both applications are assigned to a user with ID `john.doe@example.com`, and both provide full read-only access to an AWS account with ID `1234567891011`.
- two profiles in `~/.aws/config` for both applications:
```
[profile normal-app]
aws_saml_url = home/amazon_aws/XXXXXXXXXXX
role_arn = arn:aws:iam::1234567891011:role/

[profile app-with-mfa]
aws_saml_url = home/amazon_aws/YYYYYYYYYYYY
role_arn = arn:aws:iam::1234567891011:role/
```
Run results:
```
$ aws-okta -d exec normal-app -- aws ec2 describe-instances
DEBU[0000] Parsing config file /Users/john.doe/.aws/config
DEBU[0000] Using aws_saml_url from profile: normal-app
DEBU[0000] using okta provider
DEBU[0001] Failed to reuse session token, starting flow from start
DEBU[0001] Step: 1
DEBU[0001] Step: 2
DEBU[0001] Step: 3
DEBU[0004] Writing session for normal-app to keyring
DEBU[0004]  Using session HAAO, expires in 59m58.763470426s
...<successful command execution>

$ aws-okta -d exec app-with-mfa -- aws ec2 describe-instances
DEBU[0000] Parsing config file /Users/john.doe/.aws/config
DEBU[0000] Using aws_saml_url from profile: app-with-mfa
DEBU[0000] using okta provider
DEBU[0004] Failed to reuse session token, starting flow from start
DEBU[0004] Step: 1
DEBU[0005] Step: 2
DEBU[0005] Step: 3
Okta user john.doe@example.com does not have the AWS app added to their account.  Please contact your Okta admin to make sure things are configured properly.

### PR DESCRIPTION
**Pull request by deiwin**
Okta APIs don't support app-level MFA. Even the [documentation for their
own AWS CLI tool][1] says:

> Note: At this time, per-app MFA is not supported in this integration.
Only organization-level MFA is supported.

This provides an alternative to the SAML client, that works better with MFA
being enabled only for a specific Okta application. With SAML it's
difficult to get the control back to the CLI program once the user has
finished authentication. With OIDC it's easier and we can then force the OS
to focus the window from which the CLI program is run, creating a
reasonable smooth flow. It's not ideal - not as good as being able to input
the MFA information through the CLI, but the best I could come up with
given the APIs.

Setup instructions are included in the readme.

Fixes #63

[1]: https://support.okta.com/help/s/article/Integrating-the-Amazon-Web-Services-Command-Line-Interface-Using-Okta#Execution1

### CODE REVIEW

**Comment by deiwin** [comment]
> One thing: this does break API compat right?

Not at all. Everything that worked previously should keep working exactly like it did before. It only changes behavior if you specify a `oidc_app_id` in `~/.aws/config`.

**Comment by nickatsegment** [comment]
Yep, you're right, looks like it only adds public fields/methods, so that's probably fine.

**Comment by deiwin** [comment]
Just FYI. I also tried using the Early Access [_IDP-initiated Step-up Authentication_ method](https://developer.okta.com/docs/api/resources/authn#idp-initiated-step-up-authentication) which looked promising. You just specify the app ID as `audience` when making the very first `/api/v1/authn` request and the app's Sign On preferences, including app-specific MFA, will be considered. This means that the existing MFA code just works for authenticating for a specific app. However, once you finish authentication, all you get with a SUCCESS response is a link which you have to navigate to, just to start the authentication process all over again (the page has a `SAMLRequest`, not a  `SAMLResponse`). No session token or anything.

**Comment by rplute** [comment]
Hi, @deiwin, thanks for this! I set up this fork, but had to add the Client Secret of the Okta OIDC app to the Exchange method used. My understanding is that the Okta Authorization Code grant requires the Client Secret. Is this not the case or did I not set up the Okta OIDC app properly? Thanks!

**Comment by deiwin** [comment]
Hi. Did you follow the steps in the README? It's built to use PKCE and so shouldn't require the `ClientSecret`. README says:

> Create a **Native** Okta app, with *OpenID Connect* sign on method

When I do that in Okta, I get the following _Client Credentials_ section, which works as expected, without `ClientSecret`, using PKCE.
screen shot 2018-09-26 at 09 10 51

**Comment by rplute** [comment]
@deiwin, thanks for the info! I'm not an Okta admin and had an OIDC native app created for me on my behalf with _Authorization Code Flow_, but not _Authorization Code Flow with PKCE_. It's working as intended now!

**Comment by davidmarkstartapp** [comment]
Thanks @deiwin 
Works for me!

For Windows / distros this addition should work the same?

**Comment by deiwin** [comment]
Same for Linux. Everything should work except for moving the focus back to the terminal.

**Comment by davidmarkstartapp** [comment]
Do you know when this PR will be merged?

**Comment by Ngibb** [comment]
Are there still plans to integrate this?

**Comment by nickatsegment** [comment]
@Ngibb 	I haven't had time to review, and it looks that's meant some drift from master. I should have time next week. If @deiwin 	wants to update their fork, that'd be helpful. Otherwise someone else needs to own the branch/PR.

**Comment by deiwin** [comment]
@nickatsegment, I updated the branch and resolved the conflicts. I haven't tested this updated version, however, and there seems to be something off with the dependencies in CI, which I'm not seeing locally.

**Comment by nickatsegment** [comment]
I got some testing done this week. Ran into a panic that I need to dig into more.

Honestly, it might be a while before this gets merged, if it ever gets merged. It's a pretty big feature that I don't know if I'd have the time to support.

**Comment by Ngibb** [comment]
> Honestly, it might be a while before this gets merged, if it ever gets merged. It's a pretty big feature that I don't know if I'd have the time to support.

@nickatsegment Is there something I can do to support the cause? :beer: :pizza: :question:

**Comment by nickatsegment** [comment]
> @nickatsegment Is there something I can do to support the cause? 🍺 🍕 ❓

Upvote it! Upvotes tell us what's valuable to the community and would help sway a cost-benefit analysis ;)

File: lib/okta_oidc.go [hunk 1]
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (o *OktaOIDCClient) AuthenticateProfile(profileARN string, duration time.Duration) (sts.Credentials, error) {
+	// Allow the whole operation to maximally take 2 minutes. Should be enough
+	// to cover different network delays + fumbling to find an MFA device etc.
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	// Make sure early returns close all open connections etc
+	defer cancel()
+
+	provider, err := oidc.NewProvider(ctx, fmt.Sprintf("https://%s.%s", o.Organization, OktaServerDefault))
```
nickatsegment: This hard-coded default is not ideal.

nickatsegment:
@deiwin 	Looks like add assumes a SAML config :/ If you don't have one in your config it panics.

lib/okta_oidc.go
```diff
@@ -0,0 +1,249 @@
...
+func (o *OktaOIDCClient) stopRedirectServer(srv *http.Server) {
+	log.Debug("Shutting server down")
+	if err := srv.Close(); err != nil {
+		// This should be safe to ignore in a CLI program, which shuts down
+		// shortly after anyway
+		log.Debug("Failed to shut down server. Ignoring")
+	}
+}
+
+func restoreWindowFocusAfter(f func() error) error {
```
nickatsegment: How essential is this for users? Running Applescript feels gross and requires the user to acknowledge a security prompt. I'd love to cut this out.

lib/okta_oidc.go
```diff
@@ -0,0 +1,249 @@
...
+		fmt.Sprintf("tell application \"%s\" to activate", windowWithFocus)).Run()
+	if eErr, ok := err.(*exec.ExitError); ok {
+		log.Debugf("Refocusing failed. Ignoring. Error: %s", string(eErr.Stderr))
+	} else if err != nil {
+		return err
+	}
+	return fErr
+}
+
+func pkce() (string, string, error) {
```
nickatsegment: Where is this implementation from? Is it original?

lib/okta_oidc.go
```diff
@@ -0,0 +1,249 @@
...
+	codeVerifier, err := randomHex(30)
```
nickatsegment: Why 30?

lib/okta_oidc.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return "", "", err
+	}
+
+	hash := sha256.New()
```
nickatsegment: Any reason not to just use `Sum256`?

lib/okta_oidc.go
nickatsegment: And why hex?

Based on [this doc](https://www.oauth.com/oauth2-servers/pkce/authorization-reque), the string can contain `[A-Za-z0-9._~-]`, and can be 43-128 characters long

nickatsegment:
The build definitely needs fixing. It's likely you're not building in a clean room, like CI does. We do this to make sure `vendor.json` is correct.

Twitch:
> > @nickatsegment Is there something I can do to support the cause? 🍺 🍕 ❓
> 
> Upvote it! Upvotes tell us what's valuable to the community and would help sway a cost-benefit analysis ;)

Is there's anything else we can do to assist, I'd love to know. With a push to have our org using app-level MFA, I'd love to ensure we don't break any of the other workflows that our teams use aws-okta for today. I can feel the shady workarounds that will ensue in my bones already. :-)

nickatsegment:
> > > @nickatsegment Is there something I can do to support the cause? 🍺 🍕 ❓
> > 
> > 
> > Upvote it! Upvotes tell us what's valuable to the community and would help sway a cost-benefit analysis ;)
> 
> Is there's anything else we can do to assist, I'd love to know. With a push to have our org using app-level MFA, I'd love to ensure we don't break any of the other workflows that our teams use aws-okta for today. I can feel the shady workarounds that will ensue in my bones already. :-)

Haven't heard from @deiwin re: review in quite some time, and there are conflicts now. If you want to adapt the PR and submit a new one, we could go from there.

Alternatively, since this is kind of a big change with some dense security-sensitive bits, maybe a fork is in order. Maintaining a big feature with no in-house users is not something I look forward to.

deiwin:
Yeah, sorry, haven't had the time for this.

nickatsegment:
Closing due to age. Ping me if you want it reopened, or start a new PR.

PR #84, opened 2018-09-27
Add README update to show support for EKS and Kubectl.

jaredm4:
You solved a problem you didn't even know about. :) I had `aws-okta` working for `aws-cli`, but whenever I tried to access my EKS cluster with `kubectl`, it would fail since `aws-iam-authenticator` doesn't know anything about `aws-okta`.

This works so long as `aws-okta` uses ENV variables to authenticate with AWS.

Adding this info to README should increase SEO friendliness on the topic and help others like myself find it in the future as EKS gains traction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,19 @@ Global Flags:
   -d, --debug            Enable debug logging
 ```
 
+### Exec for EKS and Kubernetes
+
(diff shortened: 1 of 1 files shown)


nickatsegment:
Cool! Glad that works. I think I'd like to see this section come later, maybe between the Backends and Releasing section because I think it applies to less users than the config stuff. That said the order of the sections as it stands isn't great (config should come before

PR #85, opened 2018-10-03
Support Yubikey configured in Duo for 2FA

prasincs:
This breaks a few abstractions -- and hence warrants some explanation
The Duo MFA flow in aws-okta is predicated on something like this

1. user pings okta
2. if duo, call duo API /frame/prompt which sends a push
3. this is where we have problem.. with yubikey vs push
-- instead of async waiting, yubikey can write directly to stdin and
continue, because the initial factor call wants the passcode to be
passed in
4. Then in Status call turns out the first status call returns empty
for Push but second can wait. It's opposite for Yubikey

The next step is on how to give the users a choice. I thought of few
options and settled on a flag, where the user flow is - X method
didn't work, I'll try another one. But because of the providers
nesting and calling OktaClient, it MFADevice field had to be added to
bunch of structures, where they feel very leaky. The structure doesn't
lend itself to anything more clever currently.

So the flag I introduced is --mfa-device which defaults to phone1 so
users expecting previous behavior can continue working as
is. Additionally this introduces another environment variable
AWS_OKTA_MFA_DEVICE which can be token, phone1, phone2, whatever Duo
will accept.

Alternate and a maybe architecturally cleaner way might be to ask user
which factor they want. But picking every single time when a flag
would do seems inefficient too.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,11 @@ jobs:
           command: |
             sudo make -f Makefile.tools nfpm-debian rpmbuild-debian
       - run:
+          name: Install libusb
+          command: |
+            sudo apt update -q
+            sudo apt install -yq libusb-dev
+      - run:
           name: Make distributables
           command: |
             make -f Makefile.release dist-linux
@@ -39,11 +44,11 @@ jobs:
       - run:
           name: Install tools
           command: |
-            # this is required for the u2f to work linux & for package_cloud :/
-            sudo apt update -q
-            sudo apt install -yq ruby ruby-dev build-essential libusb-1.0.0-dev libusb-dev
-            # fixes https://askubuntu.com/questions/872399/error-failed-to-build-gem-native-extension-when-trying-to-download-rubocop
             make -f Makefile.tools github-release
+            # this is all for package_cloud :/
+            sudo apt update -q
+            sudo apt install -yq ruby ruby-dev build-essential
+            # fixes https://askubuntu.com/questions/872399/error-failed-to-build-gem-native-extension-when-trying-to-download-rubocop
             sudo gem install rake
             sudo make -f Makefile.tools package_cloud
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,11 @@ jobs:
     steps:
       - checkout
       - run:
+          name: Install linux dependencies - libusb
+          command:
+            sudo apt update -q
+            sudo apt install -yq libusb-dev libusb-1.0.0-dev
+      - run:
           name: Test
           command: |
             make test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,8 +41,8 @@ jobs:
           command: |
             make -f Makefile.tools github-release
             # this is all for package_cloud :/
-            sudo apt update -q 
-            sudo apt install -yq ruby ruby-dev build-essential
+            sudo apt update -q
+            sudo apt install -yq ruby ruby-dev build-essential libusb-1.0.0-dev libusb-dev
             # fixes https://askubuntu.com/questions/872399/error-failed-to-build-gem-native-extension-when-trying-to-download-rubocop
             sudo gem install rake
             sudo make -f Makefile.tools package_cloud
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
       - checkout
       - run:
           name: Install linux dependencies - libusb
-          command:
+          command: |
             sudo apt update -q
             sudo apt install -yq libusb-dev libusb-1.0.0-dev
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ jobs:
           name: Install libusb
           command: |
             sudo apt update -q
-            sudo apt install -yq libusb-dev
+            sudo apt install -yq libusb-dev libusb-1.0.0-dev
       - run:
           name: Make distributables
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,11 +39,11 @@ jobs:
       - run:
           name: Install tools
           command: |
-            make -f Makefile.tools github-release
-            # this is all for package_cloud :/
+            # this is required for the u2f to work linux & for package_cloud :/
             sudo apt update -q
             sudo apt install -yq ruby ruby-dev build-essential libusb-1.0.0-dev libusb-dev
             # fixes https://askubuntu.com/questions/872399/error-failed-to-build-gem-native-extension-when-trying-to-download-rubocop
+            make -f Makefile.tools github-release
             sudo gem install rake
             sudo make -f Makefile.tools package_cloud
       - run:
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -20,6 +20,11 @@ var (
 	ErrFailedToValidateCredentials = errors.New("Failed to validate credentials")
 )
 
+const (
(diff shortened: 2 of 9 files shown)


nickatsegment:
Cool stuff. I'm going to take a look and give this a spin today.

nickatsegment:
I couldn't get it to work. I get to `Press button on your hardware token:`, but pressing the button does nothing. The button isn't lighting up like it would on a u2f website.

I thought the u2f protocol needed to issue a challenge to the key or something? I'm not seeing that here.

prasincs:
Can you tell me about the setup?

I tried it on Yubikey 4 on a OSX 10.14 machine -- it registered as a
keyboard more or less, and I could just press the button. I haven't tried
in other systems. It may require doing some extra work? I'd be happy to
check it out for a different u2f device if I can get hold of one.
This is the method I used for registering the yubikey
https://duo.com/docs/yubikey


https://asciinema.org/a/rocbwsrhj8ho7VaYUYBCEdiin

prasincs:
> it's OTP, not U2F
Yeah. I was confused about that in the implementation too and existing
comments didn't help.
I can look into the U2F setup eventually but probably not in next week or
two.

Likely will be another PR :)

prasincs:
@nickatsegment so it has my "U2F works on my machine(s)" badge.. at least for OSX. Haven't figured out Linux flow yet, but the build is breaking.. I'd rather not poke around too much in circleci configs if you can take a look into that. 
And this should be classified as "very alpha quality PoC to test on"..

.circleci/config.yml
```diff
@@ -39,11 +39,11 @@ jobs:
       - run:
           name: Install tools
           command: |
-            make -f Makefile.tools github-release
-            # this is all for package_cloud :/
-            sudo apt update -q 
-            sudo apt install -yq ruby ruby-dev build-essential
+            # this is required for the u2f to work linux & for package_cloud :/
+            sudo apt update -q
+            sudo apt install -yq ruby ruby-dev build-essential libusb-1.0.0-dev libusb-dev
```
nickatsegment: I think you want to install libusb* during the `dist-linux` job, and not this one.

nickatsegment:
Oh wow, I wasn't actually suggesting you go ahead and do it. Thanks for that!

Turns out someone in Segment is working on an implementation for OTP/U2F for an internal tool, and it'd be 90% the same code. I know he's on vacation ATM, but I'd love to loop him in and maybe y'all could work together. @alberts 

Re: failing build: I can't update your branch :) I commented on the PR

prasincs:
Yeah, I kinda wanted something like this for a workshop next week at
AppSecUSA. Adding tokens was easy, so figured I'll hack something to see
how u2f works for a bit. :)

I'm happy with whatever changes @alberts thinks will fit. There's lot of
stuff in FIDO2 spec (and/or depending on what your Duo setup is like) I
decided to skip over. Also needs more testing to make u2f devices detect,
not sure if it's how I'm opening devices or something else. Happy to help
wherever I can. Disconnecting/reconnecting to get it working is okay, but
probably should be frontloaded if you know there's a u2f device to check

.circleci/config.yml
```diff
@@ -21,6 +21,11 @@ jobs:
           name: Install nfpm, rpmbuild
           command: |
             sudo make -f Makefile.tools nfpm-debian rpmbuild-debian
+      - run:
```
nickatsegment: Based on the latest test failure, looks like we need it for tests too. I should have realized that.

prasincs: thanks, it's fixed now

nickatsegment:
Cool, I'm getting somewhere with a build. Having trouble with `getTrustedFacet`. The parameter `appId` is `https://api-3bdc2f77.duosecurity.com`, which looks like it just resolves to the main Duo site. I stubbed that out and assumed `facet = appId`, and then it went through 👍 Not really clear what's supposed to happen here

prasincs:
Weird that you get that.. mine was some path for appId. I did the
`facet=appId` initially and it didn't work in my duo orgs.. so I went the
long tail approach..
the spec is pretty detailed on what's valid and what's not valid..
https://fidoalliance.org/specs/fido-uaf-v1.2-rd-20171128/fido-appid-and-facets-v1.2-rd-20171128.html#dictionary-trustedfacetlist-members

For instance, I get two ids, one for admin and one for this in my duo orgs,
but this case you're probably safe stripping the path out and using the
https://<hostname> for facet.

nickatsegment:
Yeah, I'm not 100% clear on what facets are supposed to protect against, so I'm not sure what *not* checking it loses.

nickatsegment:
Yeah, I bet because you're a Duo admin (and I'm not), you have two potential facets, and the app would have to pick between them, whereas I don't have to pick, and should just assume appId == facetId.

lib/duo.go
```diff
@@ -79,25 +127,112 @@ func (d *DuoClient) ChallengeU2f() (err error) {
...
+		prompted := false
+		timeout := time.After(time.Second * 25)
+		interval := time.NewTicker(time.Millisecond * 250)
+		facet := ""
+		facetResponse, err := d.getTrustedFacet(status.Response.U2FSignRequest[0].AppID)
+		if err != nil {
+			return fmt.Errorf("failed to get trusted facets for u2f device. Err: %s", err)
+		}
+		// TODO: do more error checks
+		facet = facetResponse.TrustedFacets[0].Ids[0]
```
nickatsegment: Maybe this is what you meant by the TODO, but shouldn't we iterate through `TrustedFacets` to check that `appId` is in there? Dunno if it'll always be 0

cmd/root.go
```diff
@@ -96,6 +104,7 @@ func init() {
 	for _, backendType := range keyring.AvailableBackends() {
 		backendsAvailable = append(backendsAvailable, string(backendType))
 	}
+	RootCmd.PersistentFlags().StringVarP(&mfaDevice, "mfa-device", "m", "phone1", "Device to use phone1, phone2 or token")
```
nickatsegment: Should include `u2f` too

prasincs:
> shouldn't we iterate through TrustedFacets to check that appId is in
there? Dunno if it'll always be 0
You're correct. We should iterate because there's no guaratee the first one
will always be what you want to match.. is this something we ask user to
choose? or look for a hostname match (ie if it's same hostname, continue,
or if it's empty or none match, throw error?)

> Should include u2f too
Yea

nickatsegment:
>  is this something we ask user to choose?

I think Okta is telling us what to expect in `UserAuth.Embedded.Factor.Embedded.Verification.Host`? Just prepend `https://`.

prasincs:
Ah cool, lets go with that!

prasincs:
Ok, I've tested and made the default for --mfa-device more explicit. It uses `UserAuth.Embedded.Factor.Embedded.Verification.Host` which should be correct anyway.

prasincs:
Hi,
 @nickatsegment @alberts  is there any chance you can look into it? I went on vacation right after and couldn't follow up until now.

nickatsegment:
@prasincs 	Sorry, I must have missed the notification. Good enough for me. Let's do it.

lib/duo.go
```diff
@@ -210,7 +428,7 @@ func (d *DuoClient) DoPrompt(sid string) (txid string, err error) {
 //
 // The function returns the auth string required for the Okta Callback if
 // the request succeeded.
-func (d *DuoClient) DoStatus(txid, sid string) (auth string, err error) {
+func (d *DuoClient) DoStatus(txid, sid string) (auth string, status StatusResp, err error) {
```
nickatsegment: This breaks an exported function signature, but we're 0.x so YOLO

PR #86, opened 2018-10-03
Use log.Debugf instead of log.Debug

gchaincl:
I'm having this output when `keyring.Get` fails:
```
DEBU[0000] couldnt get okta creds from keyring: %sunexpected end of JSON input
```

Cause `%s` is not being interpolated.


nickatsegment:
I feel like this happens a lot. It'd be cool if we had a linter check or something for this sort of thing

PR #100, opened 2018-11-29
Add ability to configure the Okta session key value that is stored in the keyring.

vivianho:
I have a specific case I want to solve here:

In addition to using aws-okta on the command line, I am using the aws-okta library for another Go application. This means that two separate applications need to access the same session stored in the keyring. The problem here is that on Mac OSX Keychain specifically, by default, items are only accessible by the application that created the Keychain.

To solve this problem, in this pull request, I am allowing the ability to configure the key to be stored in the keychain, so the separate applications can read from different items in the Keychain. Note: I haven't tested this on non-Keychain backends, and am not sure of the implications of different keys with identical labels (works fine in Keychain).

I also made some lint fixes, which I am happy to revert.

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -24,8 +24,8 @@ import (
 )
 
 const (
-	OktaServerUs = "okta.com"
-	OktaServerEmea = "okta-emea.com"
+	OktaServerUs      = "okta.com"
+	OktaServerEmea    = "okta-emea.com"
 	OktaServerPreview = "oktapreview.com"
 	OktaServerDefault = OktaServerUs
 
@@ -61,7 +61,7 @@ type OktaCreds struct {
 	Organization string
 	Username     string
 	Password     string
-	Domain 		 string
+	Domain       string
 }
 
 func (c *OktaCreds) Validate(mfaDevice string) error {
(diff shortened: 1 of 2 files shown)


nickatsegment:
I love it. Personally, I think it'd be very useful for development of aws-okta so you can avoid having to switch the permissions in the keychain.

One thing I'm not going to be able to point out in-line: please add some documentation around the feature to the README.

lib/provider.go
```diff
@@ -146,12 +146,25 @@ func (p *Provider) getSamlURL(source string) (string, error) {
 	return "", errors.New("aws_saml_url missing from ~/.aws/config")
 }
 
+func (p *Provider) getOktaSessionKey(source string) string {
+	haystack := []string{p.profile, source, "okta"}
+	for _, profile := range haystack {
+		oktaSessionKey, ok := p.profiles[profile]["okta_session_key"]
```
nickatsegment: Likewise, `okta_session_key` could be `okta_session_cookie_key`

lib/okta.go
```diff
@@ -483,6 +483,7 @@ type OktaProvider struct {
 	SessionDuration time.Duration
 	OktaAwsSAMLUrl  string
 	MFADevice       string
+	OktaSessionKey  string
```
nickatsegment: Could this have a more descriptive name? It's the key for the keychain item containing the Okta session cookie. How about `SessionCookieKey`? Even that doesn't sound descriptive enough. Maybe add a comment above to flesh it out.

README.md
```diff
@@ -110,6 +110,14 @@ role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
 # This profile uses the "integrations-auth" Okta app combined with secondary role assumption
 source_profile = integrations-auth
 role_arn = arn:aws:iam::<account-id>:role/<secondary-role-name>
+
+[profile testaccount]
+# This stores the Okta session in a separate item in the Keyring.
+# This is useful if the Okta session is used or modified by other applications
+# and needs to be isolated from other sessions. It is also useful for
+# development versions or multiple versions of awsokta running.
```
nickatsegment: `aws-okta` :) Then we're good to go

nickatsegment:
Awesome, thanks for your work @vivianho

PR #103, opened 2018-12-17
Increase timeouts to allow auth over slow connections

jayzes:
This is a pretty small change, but the issue always hamstrings me when I fly and I usually forget about it right after I land.

Currently either the TLS handshake (10s default) or subsequent connection can time out on a slow network - for example, airplane wi-fi. This makes those timeouts 60s apiece so those attempts have a better chance of getting through successfully.

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -31,6 +31,8 @@ const (
 
 	// deprecated; use OktaServerUs
 	OktaServer = OktaServerUs
+
+	Timeout = time.Duration(60 * time.Second)
 )
 
 type OktaClient struct {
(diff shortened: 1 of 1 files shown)


lib/okta.go
```diff
@@ -436,8 +436,14 @@ func (o *OktaClient) Get(method string, path string, data []byte, recv interface
 		header = http.Header{}
 	}
 
+	timeout := time.Duration(60 * time.Second)
```
nickatsegment: Extract this number out to a const

nickatsegment:
I'm trying to think of what the downside is to such a long timeout. I guess if your network is truly down, it'd take a pretty long time for aws-okta to tell you that.

Anybody else have feelings? @systemizer?

jayzes:
Excellent - just adjusted with the constant change. Let me know if you want me to do anything else!

PR #97, opened 2018-11-19
Implement config option for which MFA to use

lsowen:
A few additional changes to #73 to address code review feedback.

PR #108, opened 2018-12-27
Implement config option for which MFA to use

lsowen:
Supercedes #97/#73, and streamlines changes for this PR and #85.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,14 @@ role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
 assume_role_ttl = 12h
 ```
 
+#### Multi-factor Authentication (MFA) configuration
+
+If you have a single MFA factor configured, that factor will be automatically selected.  By default, if you have multiple available MFA factors, then you will be prompted to select which one to use.  However, if you have multiple factors and want to specify which factor to use, you can do one of the following:
+
+* Specify on the command line with `--mfa-provider` and `--mfa-factor-type`
+* Specify with environment variables `AWS_OKTA_MFA_PROVIDER` and `AWS_OKTA_MFA_FACTOR_TYPE`
+* Specify in your aws config with `mfa_provider` and `mfa_factor_type`
+
 ## Backends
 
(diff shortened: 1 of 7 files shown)


lsowen:
hi @nickatsegment, happy new year!  Wanted to see if you have some feedback on CLI vs AWS config to do the MFA settings.

nickatsegment:
@lsowen 	Hi, happy new year! Thanks for hanging in there despite our radio silence :)

Do you have (or can you imagine) a use case where you'd want different MFA methods depending on the account? For me, as a user, I only ever use 1 idP (same for every AWS account) and always pick the same method. So for my own personal use case, I'd prefer an env var because you can set it in your bash profile and not have to copy it to every aws-okta profile.

lsowen:
@nickatsegment I think whichever method we use (cli flag or aws config) I will implement an ENV variable override.

Would it be useful to put plumbing in place to allow values to be set in the `[okta]` section and overridden in the individual `[profile XYZ]` sections?  That way you could set it once, or override it if you wanted/needed.

lsowen:
@Fauzyy just rebased, updated to use the new `GetValue`, and updated the README.

This version supports:
* Setting from CLI
* Setting from environment variables
* Setting from AWS config

Fauzyy:
@lsowen this look good to me so far, I'll merge once I wrap up testing 👍 - thanks!

manishtomar:
Strangely I don't see this working after taking the latest code and installing it.
```
$ aws-okta login --mfa-factor-type push ds
INFO[0001] Select a MFA from the following list
INFO[0001] 0: OKTA (push)
INFO[0001] 1: OKTA (token:software:totp)
Select MFA method: ^C
```
Isn't it supposed to default to push instead of asking it? Did I misunderstand the documentation?

lsowen:
hi @manishtomar, while you can mix/match the methods of specifying which one you want to use (eg use environment variable and CLI switch), you have to specify both the provider and the factor-type for the config to have impact.  eg

```
aws-okta login --mfa-provider OKTA --mfa-factor-type push ds
```

manishtomar:
@lsowen Awesome. Thank you so much for the clarification and thanks a lot for this contribution!

PR #111, opened 2019-01-17
Add a general method for retrieving profile configuration values

lsowen:


lsowen:
@nickatsegment A separate PR implementing what I was talking about in https://github.com/segmentio/aws-okta/pull/108#issuecomment-453111474

If this is acceptable (and merged) I'll update #108 to use this.

nickatsegment:
Cool. I think this introduces some different behaviour? For instance, previously you could only potentially get the value from the profile, its parent, or the root `okta`. Now ancestry could be arbitrarily long, right?

(BTW, I think arbitrary ancestry chains means that one could make loops, which this code would just spin on.)

Can you please point out any such changes?

lib/config.go
```diff
@@ -65,3 +65,27 @@ func sourceProfile(p string, from profiles) string {
 	}
 	return p
 }
+
+func GetValueFromProfile(profile string, from profiles, config_key string) (string, string, error) {
```
nickatsegment: How do you feel about making this a method on the `profiles` type?

lib/config_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if found_profile != "okta" {
+		t.Error("key_a should have come from `okta`")
+	}
+
+	if found_value != "a" {
+		t.Error("The proper value for `key_a` should be `a`")
+	}
+
+	// Traversing from a child profile
```
nickatsegment: Split these test cases using `t.Run`, since they're independent

lib/provider.go
```diff
@@ -134,37 +134,34 @@ func (p *Provider) Retrieve() (credentials.Value, error) {
...
-		if ok {
-			log.Debugf("Using aws_saml_url from profile: %s", profile)
-			return oktaAwsSAMLUrl, nil
-		}
+func (p *Provider) getSamlURL() (string, error) {
+	oktaAwsSAMLUrl, profile, err := GetValueFromProfile(p.profile, p.profiles, "aws_saml_url")
+	if err != nil {
+		log.Debugf("Using aws_saml_url from profile: %s", profile)
+		return oktaAwsSAMLUrl, nil
+
```
nickatsegment: errant whitespace

lib/provider.go
```diff
@@ -134,37 +134,34 @@ func (p *Provider) Retrieve() (credentials.Value, error) {
...
-func (p *Provider) getOktaSessionCookieKey(source string) string {
-	haystack := []string{p.profile, source, "okta"}
-	for _, profile := range haystack {
-		oktaSessionCookieKey, ok := p.profiles[profile]["okta_session_cookie_key"]
-		if ok {
-			log.Debugf("Using okta_session_cookie_key from profile: %s", profile)
-			return oktaSessionCookieKey
-		}
+func (p *Provider) getOktaSessionCookieKey() string {
+
```
nickatsegment: errant whitespace

lib/provider.go
```diff
@@ -134,37 +134,34 @@ func (p *Provider) Retrieve() (credentials.Value, error) {
...
+	oktaSessionCookieKey, profile, err := GetValueFromProfile(p.profile, p.profiles, "okta_session_cookie_key")
+	if err != nil {
+		log.Debugf("Using okta_session_cookie_key from profile: %s", profile)
+		return oktaSessionCookieKey
 	}
+
```
nickatsegment: errant whitespace

lib/provider.go
```diff
@@ -134,37 +134,34 @@ func (p *Provider) Retrieve() (credentials.Value, error) {
...
 	source := sourceProfile(p.profile, p.profiles)
-	oktaAwsSAMLUrl, err := p.getSamlURL(source)
+	oktaAwsSAMLUrl, err := p.getSamlURL()
 	if err != nil {
 		return sts.Credentials{}, err
 	}
-	oktaSessionCookieKey := p.getOktaSessionCookieKey(source)
+	oktaSessionCookieKey := p.getOktaSessionCookieKey()
 
 	profileARN, ok := p.profiles[source]["role_arn"]
```
nickatsegment: Could you use your new function/method here? (And then `source` becomes dead.)

lib/config.go
lsowen: @nickatsegment the one complication is that `profiles` is not an exported type, which I think makes https://github.com/segmentio/aws-okta/pull/111/files#diff-aed93a53379775027b6616ae133e3815R57 not work?  I think it is cleaner to put it on `profiles`, but should I export the `profiles` type?

lib/provider.go
lsowen: The current logic is to use the "parent" `role_arn` to authenticate with AWS, then "pivot" into the current profile `role_arn`.  I could see a couple of different ways of handling this:

1. Current behaviour: Authenticate with immediate parent, then pivot into the current profile role
2. Recursive: Authenticate as the "root" parent (possibly grandparent, etc), then privot into each role down until the current profile
3. Something else?

Let me know your thoughts

nickatsegment: Ah yes, this seems like a slightly different usage. Just leave it as current behaviour

nickatsegment:
Documenting new behaviour: I was thinking just in the PR to start.

But I'd say just keep the behaviour the same. I think you're anticipating needs that no user has shown. We can cross that bridge later.

lsowen:
@nickatsegment it no longer searches recursively, so implements the previous behaviour.  Additionally, changed the function name to `GetValue()` (`Get` seemed like it should return a profile, so I opted for a slightly more descriptive name).

lib/config.go
```diff
@@ -57,11 +57,35 @@ func (c *fileConfig) Parse() (profiles, error) {
...
 	if conf, ok := from[p]; ok {
 		if source := conf["source_profile"]; source != "" {
 			return source
 		}
 	}
 	return p
 }
+
+func (p Profiles) GetValue(profile string, config_key string) (string, string, error) {
+	for i := 0; i < 2; i++ {
```
nickatsegment: I feel like this would be shorter and clearer if you unrolled the loop.

lib/config.go
```diff
@@ -57,11 +57,35 @@ func (c *fileConfig) Parse() (profiles, error) {
...
+	}
+
+	// Fallback to `okta` if no profile supplies the value
+	profile = "okta"
+	config_value, ok := p[profile][config_key]
+	if ok {
+		return config_value, profile, nil
+	}
+
+	return "", "", fmt.Errorf("Could not find %s in %s, parent profiles, or okta", config_key, profile)
```
nickatsegment: For `parent profiles`, it should just be source profile maybe?

lsowen:
@nickatsegment This is ready for final review and merging.

PR #114, opened 2019-01-28
Log in via the saml URL if the profile contains one

Fauzyy:
This PR changes the `login` command to prioritize the `aws_saml_url` for the target profile if it contains one. If not, it falls back to using the federated login API.

Using the federated login method is problematic if your config has multiple accounts. Any time you login to another role on a **different** account, you're greeted with a page saying you must sign out first.

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -536,3 +536,39 @@ func (p *OktaProvider) Retrieve() (sts.Credentials, string, error) {
 
 	return creds, oktaCreds.Username, err
 }
+
+func (p *OktaProvider) GetSamlLoginURL() (*url.URL, error) {
+	item, err := p.Keyring.Get("okta-creds")
+	if err != nil {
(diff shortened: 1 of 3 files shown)


lib/okta.go
```diff
@@ -536,3 +536,39 @@ func (p *OktaProvider) Retrieve() (sts.Credentials, string, error) {
 
 	return creds, oktaCreds.Username, err
 }
+
+func (p *OktaProvider) GetSamlLoginURL() (*url.URL, error) {
```
nickatsegment: We've already broken this unholy seal elsewhere, but how do you feel about capitalizing `Saml --> SAML`? For case consistency.

nickatsegment:
The unqualified word `provider` is going on my blacklist of meaningless terms :P

lib/okta.go
Fauzyy: Good call!

PR #117, opened 2019-02-04
Provide autocompletion in bash and zsh

josacar:


josacar:
I'm quite a newbie in Golang, pls gimme some feedback :)

nickatsegment:
I love this! Had been meaning to add this myself, but never got around to it. Will review today.

cmd/exec.go
```diff
@@ -20,12 +20,15 @@ var (
 	assumeRoleTTL time.Duration
 )
 
+var profiles, _ = listProfiles()
```
nickatsegment: This pollutes the global namespace and doesn't check `err`. How about this?

```

var execCmd = &cobra.Command{
	Use:       "exec <profile> -- <command>",
	Short:     "exec will run the command specified with aws credentials set in the environment",
	RunE:      execRun,
	PreRun:    execPre,
	ValidArgs: listProfileNames(mustListProfiles()),
}

func mustListProfiles() lib.Profiles {
	profiles, err := listProfiles()
	if err != nil {
		log.Panicf("Failed to list profiles: %v", err)
	}
	return profiles
}
```

cmd/autocompletion.go
```diff
@@ -0,0 +1,242 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+func runCompletionZsh(cmd *cobra.Command, args []string) error {
+	out := bufio.NewWriter(os.Stdout)
+	defer out.Flush()
+	zshInitialization := `
+__aws-okta_bash_source() {
```
nickatsegment: This is pretty ugly, but meh, shell code amirite?

cmd/autocompletion.go
```diff
@@ -0,0 +1,242 @@
...
+var autocompleteCmd = &cobra.Command{
+	Use:   "completion",
+	Short: "Output shell completion code for the given shell (bash or zsh)",
+	Long: `
+Output shell completion code for bash or zsh
+This command prints shell code which must be evaluated to provide interactive
+completion of aws-okta commands.
+
+Bash
+	$ source <(aws-okta completion bash)
```
nickatsegment: I wonder if instead of subcommands, we could do a parameter defaulting to `auto` (the current shell).

- `bash -c 'aws-okta completion'` --> bash
- `bash -c `aws-okta completion bash` --> bash
- `zsh -c `aws-okta completion` --> zsh

Seems a little bit nicer, but not at all a blocker.

cmd/autocompletion.go
```diff
@@ -0,0 +1,242 @@
...
+	$ source $(brew --prefix)/etc/bash_completion
+	$ source <(aws-okta completion bash)
+	(or, if you want to preserve completion within new terminal sessions)
+	$ echo 'source <(aws-okta completion bash)' >> ~/.bashrc`,
+	RunE: runCompletionBash,
+}
+
+var zshAutocompleteCmd = &cobra.Command{
+	Use:   "zsh",
+	Short: "List users",
```
nickatsegment: Copy error?

cmd/exec.go
josacar: Ok, this looks much better indeed!

cmd/autocompletion.go
josacar: There's still a bug open for the Zsh completion https://github.com/spf13/cobra/issues/107

cmd/autocompletion.go
josacar: I don't know how you can check this, any suggestion?

cmd/autocompletion.go
josacar: Yup, I'll fix it

cmd/autocompletion.go
nickatsegment: I was thinking of just parsing `$SHELL`. I think that's pretty standard.

josacar: $SHELL doesn't work if you change between shells, only for login shell. Maybe we can iterate it in another pr

josacar:
@nickatsegment I don't have write access so I cannot merge :) Feel free to do it if you feel so.

Thanks for maintaining this project

PR #118, opened 2019-02-05
Updated Windows docs to reflect more lessons learned and current buil…

devonbleak:
…d process


devonbleak:
Happy to help, we get a ton of use from this and glad to contribute back.

PR #126, opened 2019-02-28
Add cmd flags

devinburnette:
Firstly, I wanted to express how grateful we are for this tool. It's going to drastically change the way we do access management for the better across our engineering org. I was wondering if you think the following change is acceptable?

The first commit adds optional flags for the add command in lieu of asking for user input. We want to be able to roll this out for all engineers with an on-boarding process that's as seamless as possible. The flags will help us automate the things we already know about and only prompt for the things we don't.

The second commit is attempting to make the add command idempotent for the "okta-creds" key on a given backend. This makes it easy to run the add command every time and not get an error if the user already has their credentials populated in their keychain.

Happy to discuss our goals further and open to other ideas in accomplishing the same.

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -12,6 +12,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	organization string
+	oktaRegion   string
+	oktaDomain   string
+)
+
 // addCmd represents the add command
 var addCmd = &cobra.Command{
 	Use:   "add",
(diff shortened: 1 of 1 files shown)


vivianho:
I just realized I duplicated this PR (almost exactly too!) sorry about that: https://github.com/segmentio/aws-okta/pull/137

devinburnette:
oh cool! thanks @vivianho , no worries! great minds think alike. as long as it gets done. 🤷‍♂️

nickatsegment:
Ah, sorry @devinburnette I missed this one. #137 is merged. Thanks for your PR though.

Issue #109, opened 2019-01-11
Linux packages need to depend on libusb?


nickatsegment:
Just a hunch, but I suspect this is the case. If so, we should ensure this is documented and captured in builds

eculver:
I was able to confirm this. For Debian, you'll see something like this:

```
# pkg-config --cflags  -- libusb-1.0
Package libusb-1.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `libusb-1.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libusb-1.0' found
pkg-config: exit status 1
```

The fix is easy: `apt-get install libusb-1.0-0-dev` -- I can add some documentation and a Dockerfile to help with this.

C8H10N4OO:
Seems related, but using the packagecloud repo, an `apt-get install aws-okta` does not install `libusb-1.0` automatically.

![image](https://user-images.githubusercontent.com/13878177/58119396-bce70880-7bb7-11e9-93d7-8224fb4eaa63.png)

PR #130, opened 2019-03-06
Add Dockerfile

eculver:
This relates to a question brought up in #109. I went ahead and added a Dockerfile for executing the build in Ubuntu as a demonstration of this need for a dependency.

Should we add a  `docker build ...` to our tests to ensure that we're building against this?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-from golang as build
+from golang:1.11 as build
 
 ENV SRC github.com/segmentio/aws-okta
 ARG VERSION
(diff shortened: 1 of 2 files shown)


nickatsegment:
> Should we add a docker build ... to our tests to ensure that we're building against this?

Probably. But we can do that later.

nickatsegment:
Oh, whoops, somebody was already working on a Docker image :/ https://github.com/segmentio/aws-okta/pull/122

Dockerfile
```diff
@@ -0,0 +1,21 @@
+from golang as build
```
nickatsegment: Consider pinning this to at least Go 1.x :)

Issue #129, opened 2019-03-06
HTTP trace mode


nickatsegment:
It would be nice for debugging if we could dump all HTTP traffic.
* [Some of our errors are not detailed enough to debug](https://github.com/segmentio/aws-okta/issues/128), and without making a big error hierarchy, HTTP dumps are probably easier to deal with
* We use some undocumented janky Duo APIs that would be easier to dev against if we could just do a packet/SSL capture

1. patch the HTTP clients with a TLS key logger (follow https://golang.org/pkg/crypto/tls/#example_Config_keyLogWriter)
2. do a packet capture

Related:

* https://github.com/segmentio/aws-okta/issues/121

nickatsegment:
Meh, this wiki article is probably good enough https://github.com/segmentio/aws-okta/wiki/HTTP-debugging

PR #133, opened 2019-03-13
WIP Add TLS key log to okta and duo libs, use in `add`

nickatsegment:
This is a POC to show how TLS key logging might look, addressing #129. TLS key logging works by logging TLS session keys to a log file, which can then be used to decrypt a packet capture, even in the face of perfect forward secrecy.

- User must build with special flag: `go build -ldflags='-X cmd.UseTLSKeyLogFile=yes'`; this would be disabled in general release
- Start Wireshark packet capture
- Run `SSLKEYLOGFILE=./keylog aws-okta add`
- [Configure Wireshark to use key log](https://security.stackexchange.com/questions/35639/decrypting-tls-in-wireshark-when-using-dhe-rsa-ciphersuites/42350#42350)
- Use the filter `ssl && http` to see decrypted SSL.

I believe this would be fairly straightforward way for advanced users to be able to debug their raw HTTP traffic.

Alternatives:

# Setting up a MITM with omni-CA a la [Charles](https://www.charlesproxy.com/documentation/using-charles/ssl-certificates/)

MITM's all traffic and requires installing a root CA. Not very :lock:

# Dump all HTTP traffic to a file instead

- Probably more complicated in the code
- What dump format?

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -92,11 +92,10 @@ func add(cmd *cobra.Command, args []string) error {
 		return ErrFailedToValidateCredentials
 	}
 	keyLogFile := addTLSKeyLog(oktaClient)
-	defer func() {
-		if keyLogFile != nil {
-			keyLogFile.Close()
-		}
-	}()
+
+	if keyLogFile != nil {
+		defer keyLogFile.Close()
+	}
 
 	if err := creds.ValidateWithClient(oktaClient); err != nil {
 		log.Debugf("Failed to validate credentials: %s", err)
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -86,8 +86,19 @@ func add(cmd *cobra.Command, args []string) error {
 	// to centralize the MFA config logic
 	var dummyProfiles lib.Profiles
 	updateMfaConfig(cmd, dummyProfiles, "", &mfaConfig)
+	var oktaClient *lib.OktaClient
+	if oktaClient, err = lib.NewOktaClient(creds, "", "", mfaConfig); err != nil {
+		log.Debugf("Failed to initialize client: %s", err)
+		return ErrFailedToValidateCredentials
+	}
+	keyLogFile := addTLSKeyLog(oktaClient)
+	defer func() {
+		if keyLogFile != nil {
+			keyLogFile.Close()
+		}
+	}()
 
-	if err := creds.Validate(mfaConfig); err != nil {
+	if err := creds.ValidateWithClient(oktaClient); err != nil {
 		log.Debugf("Failed to validate credentials: %s", err)
 		return ErrFailedToValidateCredentials
 	}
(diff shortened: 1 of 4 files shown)


abraithwaite:
> MITM's all traffic and requires installing a root CA. Not very 🔒

I'd argue that MITM is much more secure.  Takes more of a conscious effort to install a root CA.

cmd/root.go
```diff
@@ -21,6 +22,12 @@ var (
 	ErrFailedToValidateCredentials = errors.New("Failed to validate credentials")
 )
 
+// if non-zero, will log TLS keys to this file
+var UseTLSKeyLogFile = "yes"
```
nickatsegment: I'll disable this by default in the real PR

nickatsegment:
> I'd argue that MITM is much more secure. Takes more of a conscious effort to install a root CA.

I dunno, the limited scope of key logging to me says more secure?

At any rate, I'd argue it takes a pretty conscious effort to
1. Compile a custom build with the env var enabled
2. Run that custom build with the env var set
3. Not notice the spew of security warnings

Plus it limits your ability to shoot yourself in the foot by only logging for this invocation. You can still have a standard installation of `aws-okta` that has logging totally disabled running at the same time, and it will still work fine and be secure.

boggsboggs:
Can we setup a docker image with aws-okta and a root CA installed? That would keep the scope of the root CA to only aws-okta, and might be able to work without needing custom code inside the application. Could be a reasonable pattern for debugging other apps as well if we can make it run smoothly for aws-okta.

nickatsegment:
> Can we setup a docker image with aws-okta and a root CA installed?
Yep, that's reasonable. You'd only be able to test the Linux version, but I can't imagine the HTTP behaviour would be that different

cmd/add.go
```diff
@@ -86,8 +86,19 @@ func add(cmd *cobra.Command, args []string) error {
...
 	updateMfaConfig(cmd, dummyProfiles, "", &mfaConfig)
-
-	if err := creds.Validate(mfaConfig); err != nil {
+	var oktaClient *lib.OktaClient
+	if oktaClient, err = lib.NewOktaClient(creds, "", "", mfaConfig); err != nil {
+		log.Debugf("Failed to initialize client: %s", err)
+		return ErrFailedToValidateCredentials
+	}
+	keyLogFile := addTLSKeyLog(oktaClient)
+	defer func() {
```
Fauzyy: I think this also works:
```
if keyLogFile != nil {
  defer keyLogFile.Close()
}
```

cmd/root.go
```diff
@@ -21,6 +22,12 @@ var (
...
+
+// SSL to be consistent with other producers, like Firefox and Chrome
+const TLSKeyLogFileEnv = "SSLKEYLOGFILE"
```
Fauzyy: I'm curious why we want this as the default? I thought we're checking for `if file != "" {` down below.

nickatsegment: We don't want it on by default. I just had it on by default in this draft. Have updated to be clearer

abraithwaite:
> Compile a custom build with the env var enabled

If it requires recompiling with a build flag then that seems like a reasonable compromise.

In any case, MITM proxies are pretty typical for debugging TLS both in normal development and in the security community.  Lots of options that don't require adding it as a global root certificate to your personal keychain.  Even that isn't too bad though since it's only trusted by you and can only be used by you to access websites you're trying to debug.

nickatsegment:
I did a write up of the two methods [here](https://github.com/segmentio/aws-okta/wiki/HTTP-debugging).

PR #134, opened 2019-03-13
POC add docker-compose config for mitmproxy

nickatsegment:
Shows how we could use mitmproxy instead of TLS key logging, as per https://github.com/segmentio/aws-okta/pull/133#issuecomment-472546135. User flow would look like:

1. `docker-compose -f mitm.docker-compose.yml up`
2. Open a browser to http://localhost:8081 to watch the capture
3. `docker exec -it aws-okta_aws-okta_1 aws-okta add` (e.g.)

Requires a fair amount of docker faffery that I'm not loving, but seems not too bad.


mitm.docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+  aws-okta:
+    build:
+      context: .
+      target: debug
+    entrypoint: ['/bin/sh', '-c']
+    command:
+      - |
+        ln -s /mnt/mitmcerts/mitmproxy-ca-cert.pem /usr/local/share/ca-certificates/mitmproxy.crt
+        update-ca-certificates
+        while true; do sleep 100; done
```
Fauzyy: What's this `sleep` doing?

mitm.docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+    environment:
+      HTTP_PROXY: 'mitmproxy:8080'
```
Fauzyy: handy

Dockerfile
```diff
@@ -1,21 +1,30 @@
...
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+    libusb-1.0-0 \
+    ca-certificates
+ENTRYPOINT ["/usr/local/bin/aws-okta"]
 
-FROM scratch
-COPY --from=build /build/aws-okta /
+FROM scratch as production
+COPY --from=build /aws-okta /aws-okta
```
Fauzyy: do we also want to copy this to `/usr/local/bin`?

Dockerfile
```diff
@@ -1,21 +1,30 @@
...
     git
 
-RUN CGO_ENABLED=1 go build -o aws-okta -ldflags="-X main.version=$VERSION" ${SRC}/cmd
+WORKDIR /src
+COPY . .
+
+RUN go mod download 
+RUN CGO_ENABLED=1 GOOS=linux go build -o /aws-okta -ldflags="-X main.version=$VERSION"
+
+FROM golang:1.12 as debug
```
eculver: I'm not sure I understand why this layer is here? The only artifacts are from `apt-get`?

mitm.docker-compose.yml
nickatsegment: Just sleeping forever. The idea being to give the user a container they can exec into that already has the cert and proxy configured.

Dockerfile
nickatsegment: Sorry, there's a bunch of semi-related Dockerfile stuff in here that's distracting. For a scratch container, you don't need `/usr/local/bin`, but for a container you're going to exec into, that's based on Debian, you would.

Dockerfile
nickatsegment: Yeah, it's a little weird. We use `debug` as the build target in the docker-compose.yml. I suppose a separate Dockerfile might be less confusing.

nickatsegment: I think there's a bunch of stuff that needs cleanup here, and I was trying to do too much at once, and it's clouded the main purpose of the PR

nickatsegment:
I did a write up of the two methods [here](https://github.com/segmentio/aws-okta/wiki/HTTP-debugging).

PR #136, opened 2019-03-28
remove apt cache layer

smiller171:
Even in a build stage it can only hurt to keep the build cache around in its own layer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-from golang:1.11 as build
+FROM golang:1.11 as build
 
 ENV SRC github.com/segmentio/aws-okta
 ARG VERSION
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,21 +1,21 @@
...
-RUN apt-get install --no-install-recommends -y \
+RUN apt-get update && apt-get install --no-install-recommends -y \
     libusb-1.0-0-dev \
     ca-certificates \
     build-essential \
     git
 
 RUN CGO_ENABLED=1 go build -o aws-okta -ldflags="-X main.version=$VERSION" ${SRC}/cmd
 
+##################################################
```
nickatsegment: Not a huge fan of this

smiller171:
Done

PR #137, opened 2019-03-28
Added CLI arguments for aws-okta add

vivianho:
- Added CLI arguments for Okta organization (deprecated?), Okta region, Okta domain and Okta username. Reason: We would rather give our employees a single simple command to run already configured with our company's variables, rather than tell them "when you see this prompt, type this"
- Updated oktaDomain example value to `["+organization+".okta.com]` because after reading the code it seems like this is actually the format that is derived, not `"+region+".okta.com`.
- Included a log line for the oktaDomain to verify this
- Indentation fixes

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -83,10 +83,6 @@ func add(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// Set calculated Okta domain if left blank
-	if oktaDomain == "" {
-	}
-
 	if username == "" {
 		username, err = lib.Prompt("Okta username", false)
 		if err != nil {
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -12,6 +12,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	organization string
+	oktaRegion   string
+	oktaDomain   string
+)
+
 // addCmd represents the add command
 var addCmd = &cobra.Command{
 	Use:   "add",
@@ -21,6 +27,10 @@ var addCmd = &cobra.Command{
 
 func init() {
 	RootCmd.AddCommand(addCmd)
+	addCmd.Flags().StringVarP(&organization, "organization", "", "", "Okta organization name")
+	addCmd.Flags().StringVarP(&oktaRegion, "okta-region", "", "", "Okta region (us, emea, preview)")
+	addCmd.Flags().StringVarP(&oktaDomain, "okta-domain", "", "", "Okta domain (e.g. <orgname>.okta.com)")
+	addCmd.Flags().StringVarP(&username, "okta-username", "", "", "Okta username")
 }
 
 func add(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


cmd/add.go
```diff
@@ -45,30 +55,39 @@ func add(cmd *cobra.Command, args []string) error {
...
 	}
 	if oktaRegion == "" {
 		oktaRegion = "us"
 	}
 
-	oktaDomain, err := lib.Prompt("Okta domain ["+oktaRegion+".okta.com]", false)
-	if err != nil {
-		return err
+	if oktaDomain == "" {
+		oktaDomain, err = lib.Prompt("Okta domain ["+organization+".okta.com]", false)
```
nickatsegment: I've been meaning to look into this. I think the current implementation is incorrect. I'm pretty sure it should be something like

```
tld := lib.GetOktaDomain(region)
fmt.Sprintf("[%s.%s]", organization, tld)
```

getOktaDomain is in `lib` and is not exported; I'd be fine if you exported it.

cmd/add.go
```diff
@@ -45,30 +55,39 @@ func add(cmd *cobra.Command, args []string) error {
...
+		organization, err = lib.Prompt("Okta organization", false)
+		if err != nil {
+			return err
+		}
 	}
 
-	oktaRegion, err := lib.Prompt("Okta region ([us], emea, preview)", false)
-	if err != nil {
-		return err
+	if oktaRegion == "" {
```
nickatsegment: I think there's some edge cases that are going to happen here. Namely domain can be derived from region+org, and region+org can be derived from domain. You end up getting prompted for values that are implied.

```
$ ./aws-okta add --okta-domain example.okta.com
Okta organization: example?

Okta region ([us], emea, preview): us?
```

nickatsegment:
Thanks @vivianho!

FYI, organization is deprecated in OktaCreds/Client because it's superfluous. Domain is all it really cares about. There's compatibility code in there, but we should avoid using it.

cmd/add.go
```diff
@@ -21,6 +27,10 @@ var addCmd = &cobra.Command{
 
 func init() {
 	RootCmd.AddCommand(addCmd)
+	addCmd.Flags().StringVarP(&organization, "organization", "", "", "Okta organization name")
+	addCmd.Flags().StringVarP(&oktaRegion, "okta-region", "", "", "Okta region (us, emea, preview)")
```
nickatsegment: What do you think about dropping `okta` from the flag? So it's just `--region`,`--domain`, `--username`?

cmd/add.go
nickatsegment: It might be too much trouble since you'd have to mock out `lib.Prompt` probably, but if you feel like it, a table of tests would be cool. Like "if domain flag is set then expect no region, org, or domain prompt" etc.

cmd/add.go
vivianho: Yep sounds good, `region` could be confused with AWS region but if you don't think that's a problem I can just remove okta from all fields.

vivianho:
@nickatsegment your comments make sense to me. I'm trying to think about the simplest way forward here. Options:

1. Command line only takes in `--domain`, and organization/region is calculated from this.
2. Command line only takes in `--region` and `--organization` and domain is calculated from this.

It seems like 2) would be simpler (with the code that is already provided) however 1) seems to be the desired approach given that organization is deprecated?

cmd/add.go
vivianho: you are right, the prompt never showed the right `tld` if the user is not picking the us region. The code resolves it fine in the end if left blank, so it's just a UI fix.

vivianho:
I refactored with some new changes.

nickatsegment:
I like this approach. Great job.

PR #145, opened 2019-04-18
fix: idempotent publishes

nickatsegment:
So we can rebuild in the face of temporary failure. 

For example, this publish failed https://circleci.com/gh/segmentio/aws-okta/452, then this attempt to rerun failed https://circleci.com/gh/segmentio/aws-okta/454

Using this branch and a prerelease tag:
- [initial run](https://circleci.com/workflow-run/84c493b1-e4f8-4128-989e-a714e00fb5b7)
- [rebuild](https://circleci.com/workflow-run/78ecb644-9c7e-4aac-9a84-2123d0b1fb1b)

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -5,11 +5,17 @@
 
 include Makefile
 
+# use bash so we can use set -o pipefail
+SHELL := /bin/bash
+
 # set --pre-release if not tagged or tree is dirty or there's a `-` in the tag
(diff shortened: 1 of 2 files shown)


nickatsegment:
I could have explained it better.

For releases to Github, if it gets an error creating the release, it just does an `info` call to double check that it does exist.

For releases to Packagecloud, it does a "yank" aka delete first (that never exits 0), then pushes.

PR #150, opened 2019-05-12
feat: Add `env` command that prints a sourceable configuration to STDOUT.

claudenm:
This PR adds a new command, `aws-okta shell` which prints to stdout sourceable commands that set environment variables to the relevant aws profile.

The output looks like this:

```bash
./aws-okta shell test
```
```bash
export AWS_ACCESS_KEY_ID=FOO
export AWS_SECRET_ACCESS_KEY=BAR
export AWS_OKTA_PROFILE=test
export AWS_DEFAULT_REGION=us-east-1
export AWS_REGION=us-east-1
export AWS_SESSION_TOKEN=[LONG TOKEN]
export AWS_SECURITY_TOKEN=[LONG TOKEN]
```

and I use it like this:

```bash
source <(aws-okta shell test)
```
The inspiration for this was to provide an in-tool command to easily switch between AWS profiles, so that a user can use the aws-cli without have to run it through the `aws-okta exec` command.  I'm switching accounts multiple times a day, and it's useful to just have the appropriate credentials in my environment for a period of time.  Currently, this is being solved several different ways at my current company:

1. Myself and others have an executable that is a wrapper for `aws-okta exec [PROFILE] -- /bin/zsh`.  The downside to this is I have to fork a new shell, pay the startup cost for the new shell, and have a bunch of subshells running.

1. Anther colleague has a script that he runs once a day that takes each profile, grabs credentials and dumps them to his `~/.aws/credentials` file.

1. Another solution is an alias to a bash function:

```bash
_okta_aws () {
	aws-okta exec [PROFILE] -- aws "$@"
}
```

I'd imagine other people have other solutions.  While there is nothing wrong with the above solutions (or any other solution for that matter)! I thought it would be nice to centralize it in the tool.

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/99designs/keyring"
+	"github.com/alessio/shellescape"
 	analytics "github.com/segmentio/analytics-go"
 	"github.com/segmentio/aws-okta/lib"
 	"github.com/spf13/cobra"
@@ -94,18 +95,18 @@ func envRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	fmt.Printf("export AWS_ACCESS_KEY_ID=%q\n", creds.AccessKeyID)
-	fmt.Printf("export AWS_SECRET_ACCESS_KEY=%q\n", creds.SecretAccessKey)
-	fmt.Printf("export AWS_OKTA_PROFILE=%q\n", profile)
+	fmt.Printf("export AWS_ACCESS_KEY_ID=%s\n", shellescape.Quote(creds.AccessKeyID))
+	fmt.Printf("export AWS_SECRET_ACCESS_KEY=%s\n", shellescape.Quote(creds.SecretAccessKey))
+	fmt.Printf("export AWS_OKTA_PROFILE=%s\n", shellescape.Quote(profile))
 
 	if region, ok := profiles[profile]["region"]; ok {
-		fmt.Printf("export AWS_DEFAULT_REGION=%q\n", region)
-		fmt.Printf("export AWS_REGION=%q\n", region)
+		fmt.Printf("export AWS_DEFAULT_REGION=%s\n", shellescape.Quote(region))
+		fmt.Printf("export AWS_REGION=%s\n", shellescape.Quote(region))
 	}
 
 	if creds.SessionToken != "" {
-		fmt.Printf("export AWS_SESSION_TOKEN=%q\n", creds.SessionToken)
-		fmt.Printf("export AWS_SECURITY_TOKEN=%q\n", creds.SessionToken)
+		fmt.Printf("export AWS_SESSION_TOKEN=%s\n", shellescape.Quote(creds.SessionToken))
+		fmt.Printf("export AWS_SECURITY_TOKEN=%s\n", shellescape.Quote(creds.SessionToken))
 	}
 
 	return nil
--- a/cmd/shell.go
+++ b/cmd/shell.go
@@ -1,0 +1,112 @@
+package cmd
+
+import (
+	"fmt"
+
+	"os"
+	"time"
+
+	"github.com/99designs/keyring"
+	analytics "github.com/segmentio/analytics-go"
+	"github.com/segmentio/aws-okta/lib"
+	"github.com/spf13/cobra"
+)
+
+// shellCmd represents the shell command
+var shellCmd = &cobra.Command{
+	Use:       "shell <profile>",
+	Short:     "shell writes a sourceable export statements to set environment variables for the current shell for the specified profile",
+	RunE:      shellRun,
+	Example:   "source <$(aws-okta shell test)",
+	ValidArgs: listProfileNames(mustListProfiles()),
+}
+
+func init() {
+	RootCmd.AddCommand(shellCmd)
+	shellCmd.Flags().DurationVarP(&sessionTTL, "session-ttl", "t", time.Hour, "Expiration time for okta role session")
+	shellCmd.Flags().DurationVarP(&assumeRoleTTL, "assume-role-ttl", "a", time.Hour, "Expiration time for assumed role")
+}
+
+func shellRun(cmd *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return ErrTooFewArguments
+	}
+
+	profile := args[0]
+	config, err := lib.NewConfigFromEnv()
+	if err != nil {
+		return err
+	}
+
+	profiles, err := config.Parse()
+	if err != nil {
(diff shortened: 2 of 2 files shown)


svanharmelen:
@claudenm I was just about to start working on a command like this as I also found all kinds of solutions (more or less similar to the examples you gave) being created by different people in our organization.

The only change that I would like to make is that it also actually sources the env vars and that it adds an env var (`AWS_OKTA_PROFILE`) that can be used to show what profile is currently being used in your prompt (`PS1`).

If you think that makes sense, I could add a PR to your branch that adds/changes the logic a bit to accomplish that. And maybe we should consider calling the command `source` in that case? So you only have to type `aws-okta source test` and you done.

What do you think? I would really love if we could combine those ideas in your PR, but if it doesn't fit your goal then I can either fork this repo or create another PR for adding a `source` command. But it feels like we are both trying to solve the same problem right?

svanharmelen:
I noticed a couple of colleagues were doing the exact same thing (also each using their own script/logic), so I was just about to start on a very similar PR. So thanks @claudenm and lets hope it will be merged shortly!

claudenm:
@svanharmelen Thanks!  I'm also happy to take a PR against my branch if you have any improvements you'd like.

svanharmelen:
@nickatsegment I'm really curious if this is something that you are willing to merge?

nickatsegment:
Yeah, sweet. Generally I'd be fine leaving this up to aliases and what not, but meh, seems harmless, and you've done a good job @claudenm 

I wonder if commands should be quoted and/or escaped? Not sure how careful we should be here.

eculver:
I agree with @nickatsegment; I'm not opposed to having this around. One thing to keep in mind is that one way or the other the credentials will eventually expire and you have to re-run it which is just another annoying thing. The other side-effect of this is that it's not immediately obvious which profile is being used for a given invocation which would lead people to always use the most permissive profile they have.

Just some thoughts. What we ultimately want is a way for these credentials to last longer or to be negotiated for us in the background somehow. Maybe an `aws-okta agent` as @nickatsegment may have mentioned offline ?? 

Anyways, thanks for contributing @claudenm 👍

claudenm:
@eculver,  I’m happy (at least to attempt) to write some kind of prompt indicator to say what profile you’re using.  Haven’t done that before, and would be good exercise.

svanharmelen:
I use `zsh` and have this in my `zshrc` to show what profile I'm currently using:

```
PROMPT='${ret_status} %{$fg[cyan]%}%c%{$reset_color%} $(aws_prompt_info)$(git_prompt_info)'
eval orange='$FG[214]'

# Show which AWS profile is used
aws_prompt_info() {
  if [[ -n $AWS_OKTA_PROFILE ]]; then
    echo "%{$fg_bold[blue]%}aws:(%{$orange%}$AWS_OKTA_PROFILE%{$fg[blue]%})%{$reset_color%} "
  fi
}
```

But I think it makes sense to let people create or tweak their own prompts as there are so much different shells and different themes. Having the `AWS_OKTA_PROFILE` env var is (IMHO) all that one needs to be able to do that.

> I wonder if commands should be quoted and/or escaped? Not sure how careful we should be here.

I'm not sure if I understand which commands you are talking about here @nickatsegment? If you are talking about the export commands, then I don't think that is needed as I never quoted or escaped any of my AWS env vars before and never saw other people doing that either.

Is there a specific use case or potential issue that you where thinking of?

nickatsegment:
> Is there a specific use case or potential issue that you where thinking of?

A bit of a contrived example, but a profile heading like
```
[profile ;cat /etc/passwd]
...
```
Gets parsed the way you'd think. Then the script includes
```
export AWS_OKTA_PROFILE=;cat /etc/passwd
```
It'd be pretty hard for this to be a real exploit, but you can see how it could break. Like I doubt people generally have profile names with shell characters in them, but who knows.

claudenm:
Thanks @svanharmelen for adding the escaping.  @nickatsegment, mind merging?

Thanks!

Claude

svanharmelen:
And releasing :wink:

cmd/env.go
```diff
@@ -94,18 +94,18 @@ func envRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	fmt.Printf("export AWS_ACCESS_KEY_ID=%s\n", creds.AccessKeyID)
-	fmt.Printf("export AWS_SECRET_ACCESS_KEY=%s\n", creds.SecretAccessKey)
-	fmt.Printf("export AWS_OKTA_PROFILE=%s\n", profile)
+	fmt.Printf("export AWS_ACCESS_KEY_ID=%q\n", creds.AccessKeyID)
```
nickatsegment: This will quote with go rules, not bash rules. For instance, `$` is not escaped; if it's `$(cat /etc/passwd)`, that'd get subbed in.

Maybe just use this library: https://github.com/alessio/shellescape

claudenm: Good point @nickatsegment -- changed in 9a3ce05

svanharmelen: I must admit that I don’t see how this can ever be a real issue, as users need to create their own config.

So the values that can be in these export commands are values that are either created locally by the users themselves or come from AWS (after you have authenticated there).

claudenm: @nickatsegment not sure why the build failed-- it builds fine locally for me, can you kick it off again?

svanharmelen: @claudenm see the PR that I opened on your branch. The build pipeline uses `govendor` and the dependency that you introduced is not in recorded in the `vendor.json` (which is used by `govendor`).

claudenm: @svanharmelen -- thanks!  I'm new to go.... 😊

svanharmelen: No problem at all!

nickatsegment: > I must admit that I don’t see how this can ever be a real issue, as users need to create their own config.
> 
> So the values that can be in these export commands are values that are either created locally by the users themselves or come from AWS (after you have authenticated there).

Yep, you're totally right that the risk seems pretty remote. However, considering we're dealing with security-sensitive stuff here, it seems prudent to take the extra step. Thanks for doing it.

svanharmelen:
Thanks for merging this one @nickatsegment! Is there a planned timeline for releasing a new version that contains the new command?

nickatsegment:
> Thanks for merging this one @nickatsegment! Is there a planned timeline for releasing a new version that contains the new command?

"Planned"? Ahahahaha.

Hot off the presses! https://github.com/segmentio/aws-okta/releases/tag/v0.21.0

svanharmelen:
Well maybe you wanted to get some other stuff in before cutting a release. You never know right :wink:

Thanks!

PR #151, opened 2019-05-17
Contribute more like chamber

nickatsegment:
I'll apply these labels with labeler. Will need to rename `wontfix` -> `stale`


.github/labels.yml
```diff
@@ -0,0 +1,26 @@
+# Scanned and autogenerated by https://github.com/tonglil/labeler
+---
+repo: segmentio/chamber
```
Fauzyy: should this be `segmentio/aws-okta`?

Issue #152, opened 2019-05-24
extend the env command to support different shells


andy-v-h:
Some of us in the community use fish 🙃 
#150 introduced the `env` command and it sure would be handy if `aws-okta env` would print env variables in a fish-friendly format. some thing like
```go
func printExport(varName, varValue string) {
	exportString := "export %s=%s\n"
	myShell, hasShell := os.LookupEnv("SHELL")
	if hasShell && strings.Contains(myShell, "fish") {
		exportString = "set -x %s %s\n"
	}
	fmt.Printf(exportString, varName, varValue)
}
```

nickatsegment:
Sure, that'd be fine. Would happily accept a PR :)

PR #153, opened 2019-05-29
Support printing in fish format

andy-v-h:
implements #152

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/99designs/keyring"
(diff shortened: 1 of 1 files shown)


cmd/env.go
```diff
@@ -95,18 +105,19 @@ func envRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	fmt.Printf("export AWS_ACCESS_KEY_ID=%s\n", shellescape.Quote(creds.AccessKeyID))
-	fmt.Printf("export AWS_SECRET_ACCESS_KEY=%s\n", shellescape.Quote(creds.SecretAccessKey))
-	fmt.Printf("export AWS_OKTA_PROFILE=%s\n", shellescape.Quote(profile))
+	printExport("AWS_ACCESS_KEY_ID", shellescape.Quote(creds.AccessKeyID))
```
nickatsegment: Move the Quote into the function to reduce repetition.

nickatsegment: Actually, I don't know fish that well. Will the `shellescape` package still work?

PR #165, opened 2019-06-19
enable prompt if role not provided

switj:
If the `role_arn` is provided, attempt to assume that role. No change to
current behaviour.

If `role_arn` is empty or not present in the profile and there is more than one
role then display the full list of roles the user can assume and prompt
them to choose the role to assume. The choice will be sticky for the
rest of the session for that profile. Whne the session ends they will be
prompted again to choose a role.

### DIFF
--- a/lib/utils.go
+++ b/lib/utils.go
@@ -3,7 +3,9 @@ package lib
 import (
 	"encoding/base64"
 	"encoding/xml"
+	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/segmentio/aws-okta/lib/saml"
@@ -11,9 +13,22 @@ import (
 	"golang.org/x/net/html"
 )
 
-//TODO: Move those functions into saml package
-
 func GetRoleFromSAML(resp *saml.Response, profileARN string) (string, string, error) {
+
+	roles, err := GetAssumableRolesFromSAML(resp)
+	if err != nil {
+		return "", "", err
+	}
+	role, err := GetRole(roles, profileARN)
+	if err != nil {
+		return "", "", err
+	}
+	return role.Principal, role.Role, nil
+}
+
+func GetAssumableRolesFromSAML(resp *saml.Response) (saml.AssumableRoles, error) {
+	roleList := []saml.AssumableRole{}
+
 	for _, a := range resp.Assertion.AttributeStatement.Attributes {
 		if strings.HasSuffix(a.Name, "SAML/Attributes/Role") {
 			for _, v := range a.AttributeValues {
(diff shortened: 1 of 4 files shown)


switj:
Hello `aws-okta` Admins,

this was a change that we needed and wanted to submit an upstream PR to see if this would be a useful change for everyone.

the new exec flow would look something like this:
```
$ aws-okta exec okta -- aws help
INFO[0010] Requesting MFA. Please complete two-factor authentication with your second device
INFO[0010] Select a MFA from the following list
INFO[0010] 0: GOOGLE (token:software:totp)
INFO[0010] 1: FIDO (u2f)
Select MFA method: 0

Enter MFA Code: <mfa_code>

0 - arn:aws:iam::<acct_id>:role/role1
1 - arn:aws:iam::<acct_id>:role/role2
2 - arn:aws:iam::<acct_id>:role/role3
3 - arn:aws:iam::<acct_id>:role/role4
Select Role to Assume: 1
```
This is a change from:
```
$ aws-okta exec okta -- aws help
Source profile must provide `role_arn`
```

In both cases I'm using the default `okta` profile that is created during setup. The existing behaviour of having a profile that includes `role_arn` is unchanged.

bazimov:
Love this feature, if it can be implemented. ❤️

a6sondhi:
> Looks sensible, except for the bonus refactoring :P

Fixed up the commit to remove the extra refactoring. Let me know what you think.

@nickatsegment

lib/utils.go
```diff
@@ -3,16 +3,18 @@ package lib
...
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/segmentio/aws-okta/lib/saml"
 	"golang.org/x/net/html"
 )
 
-//TODO: Move those functions into saml package
+func GetAssumableRolesFromSAML(resp *saml.Response) (saml.AssumableRoles, error) {
```
nickatsegment: If possible, can you restore the function `GetRoleFromSAML` and have it call these two methods? We should avoid breaking API compatibility

lib/utils.go
```diff
@@ -32,20 +34,67 @@ func GetRoleFromSAML(resp *saml.Response, profileARN string) (string, string, er
...
+	// if the user only has one role assume that role without prompting.
+	if len(roleList) == 1 {
+		return roleList[0], nil
+	}
+
+	for i, arole := range roleList {
+		fmt.Printf("%d - %s\n", i, arole.Role)
+	}
+
+	i, err := Prompt("Select Role to Assume", false)
```
nickatsegment: For the record, prompting in lib code is gross. But we already have a precedent, so I'll let it slide :)

lib/utils.go
switj: Understood, I'll try to make the change this week.

switj: @nickatsegment `GetRoleFromSAML` restored.

nickatsegment:
> @nickatsegment Mind merging this? Or are you waiting on more approvals? 🙂

Thanks for pinging me. I do need a second pair of eyes. Looks harmless but this codebase is getting hairy and hard to follow.

nickatsegment:
@switj I merged your other PR first because I feel more sure about it. Please rebase :)

switj:
@nickatsegment rebase done. thanks for merging the other PR

Shivang44:
@nickatsegment @Fauzyy @switj Thanks so much for the quick response! What's usually the process for getting a new version into homebrew? Do we update this? https://github.com/Homebrew/homebrew-core/blob/master/Formula/aws-okta.rb

nickatsegment:
> @nickatsegment @Fauzyy @switj Thanks so much for the quick response! What's usually the process for getting a new version into homebrew? Do we update this? [Homebrew/homebrew-core:Formula/aws-okta.rb@`master`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/aws-okta.rb) [](/Homebrew/homebrew-core/blob/HEAD@{2019-09-24T20:12:53Z}/Formula/aws-okta.rb)

Usually we just cut a release and a magical fairy does the rest :D I bet they'd accept a PR from you

I'll cut a release now

PR #172, opened 2019-07-02
Remove govendor; only support go modules

nickatsegment:
Maintaining both and keeping them in sync is becoming painful.


nickatsegment:
^ I'm working with our Security team to reconfigure snyk

Issue #146, opened 2019-04-22
Collapse session tokens to single keychain item?


nickatsegment:
@boggsboggs had the idea to collapse all session tokens into a single keychain item. Currently, whenever we upgrade, keychain needs to reprompt for a password to reauthorize the new binary, for each profile, which for us is in the 10s.

nickatsegment:
I'm concerned about race conditions between two `aws-okta` processes interleaving reads and writes.

- a1: `aws-okta exec dev` reads keychain item, finds token dev=dev1; needs refresh
- a2: `aws-okta exec stage` reads keychain item; finds token stage=stage1; needs refresh
- a2: refreshes stage=stage2; rewrites entire keychain item
- a1: refreshes dev=dev2; rewrites entire keychain item, including _stage=stage1_

I'm not sure the repercussions, but it feels bad

boggsboggs:
This is all cache data, so it's not an error to throw is away, only a performance degradation. When two processes collide, they'll both still succeed using their in memory copy of the data. The next time one of them runs, it'll need to go through a longer auth flow than it otherwise would have.

I wouldn't worry about the interleaving. I think, the user needing to 2FA one additional time or wait an extra 500ms is less of a problem than requiring unnecessary password entries.

nickatsegment:
I'm going to get this in, but behind a feature flag.

Wondering how we're going to do migration.

Should we worry about cleaning up the old format keychain items? Should that happen automatically?

Issue #173, opened 2019-07-05
Add analytics for session cache hit/miss


nickatsegment:
- cache hit
- cache hit and expiration is more or less than requested
- cache miss

I believe the session cache implementation KeyringSessions is broken: I suspect it misses 100% of the time. If we gathered some analytics, we could see how bad this is.

PR #174, opened 2019-07-06
One keyring item for session

nickatsegment:
Fixes: #146

Partially addresses #173 

Adds a `--session-cache-single-item` flag and env var to opt into a new session cache that stores all sessions in a single keyring item. The benefit is fewer auth prompts from macOS Keychain. On every upgrade (new aws-okta binary), the user needs to reauth 3 + N (number of sessions) times, but with this new cache, this is always just 3+1.

This refactors the `KeyringSessions` type into a session cache store and a keying function interface so that we can reuse the existing (bizarre) keying function with a new storage engine.

I might propose in another PR that we use a better keying function. I'm still trying to understand what we want from this cache.


lib/provider.go
```diff
@@ -91,44 +111,66 @@ func (p *Provider) Retrieve() (credentials.Value, error) {
...
+		ProfileName: source,
+		ProfileConf: profileConf,
+		Duration:    p.SessionDuration,
+	}
+
+	var cachedSession *sessioncache.Session
+	cachedSession, err := p.sessions.Get(key)
 	if err != nil {
-		session, err = p.getSamlSessionCreds()
+		creds, err := p.getSamlSessionCreds()
```
systemizer: small nit: using the same variable name in this scope and the outer scope _might_ confuse a future engineer

lib/provider.go
```diff
@@ -59,13 +64,18 @@ func (o ProviderOptions) ApplyDefaults() ProviderOptions {
 	return o
 }
 
+type SessionCacheInterface interface {
```
systemizer: awesome 👐

internal/sessioncache/key_orig.go
```diff
@@ -0,0 +1,39 @@
...
+// Key returns a key for the keyring item. This is a string containing the source profile name,
+// the profile name, and a hash of the duration
+//
+// this is a copy of KeyringSessions.key and should preserve behavior, *except* that it assumes `profileName`
+// is a valid and existing profile name
+func (k OrigKey) Key() string {
+	// nick: I don't understand this at all. This key function is roughly:
+	// sourceProfileName + hex(md5(duration + json(profileConf)))
+	// - why md5?
+	// - why the JSON of the whole profile? (especially strange considering JSON map order is undetermined)
```
Fauzyy: Do we also want another implementation of the `Key` interface in this PR to play around with, or maybe that should come later

nickatsegment:
I'm gonna add some simple tests.

internal/sessioncache/key_orig.go
nickatsegment: In the new tests, I've got a thing called `fixedKey` that just returns the same string.

If you're thinking about actual, potentially useful ones, I think we'd need to understand what this cache is supposed to accomplish first ;)

Issue #189, opened 2019-07-25
Suggestion: Write env output to aws credentials file


jbialy:
AWS-okta support the `env` command which prints the export commands that can be used to set AWS environment vars. I think that it would also be useful to have an option that appends the access, secret and token values to `~/.aws/credentials` under a given [profile]? 

Perhaps something like:

```
$ aws-okta env --save my-aws
```

```
$ cat ~/.aws/credentials
[my-aws]
aws_access_key_id = ...
aws_secret_access_key = ...
aws_session_token = ...
```

What are your thought about this?

nickatsegment:
I think a different command to `env` would be better, maybe `export` or something.

That said, I feel like this can be accomplished with a pretty simple shell script:
```
aws-okta exec profile -- sh -c 'printf "[my-aws]
aws_access_key_id = \"$AWS_ACCESS_KEY_ID\"
aws_secret_access_key = \"$AWS_SECRET_ACCESS_KEY\"
aws_session_token = \"$AWS_SESSION_TOKEN\""' >> ~/.aws/credentials
```

Maybe with a few more thumbs up I'd feel better about it

fpmoles:
I feel a more appropriate use of time, especially consider the perceived use case, would be an example of how to use this within another program.

bazimov:
Yeah if we could write it into a credentials file as a profile would be awesome!

jbialy:
@nickatsegment The bash script is a good suggestion. I'm going to make use of it in a form of an alias that takes care of the logic 😄

jbialy:
> I feel a more appropriate use of time, especially consider the perceived use case, would be an example of how to use this within another program.

@fpmoles Lots of cli utilities can use `AWS_PROFILE` in order to determine where to get the AWS credentials. Having the option of storing multiple profiles in the credentials file at the same provides flexibility when working with multiple AWS accounts. Also, having to prefix commands with `aws-okta exec profile -- command` in order to provide an environment is great but also a bit limiting since you lose that environment once the command exits. Saving credentials allows you to use your shell freely with the ability to access the various AWS_PROFILEs that you have access to!

fpmoles:
> > I feel a more appropriate use of time, especially consider the perceived use case, would be an example of how to use this within another program.
> 
[...]

PR #194, opened 2019-08-16
Add export command to copy tokens to credentials file

apechimp:
This resolves https://github.com/segmentio/aws-okta/issues/189. Apologies for not getting consent via an issue before submitting the PR: the workarounds in that issue didn't work for me and this seemed to be the most straightforward path to get this to work for my scenario.

This adds a new command, `export` which will write credentials to the credentials file for the profile which was passed into the command.


toddlipcon:
I'd find this useful as well. My use case is running some integration tests from a Java IDE, where the integration tests require AWS privileges. I can invoke the IDE from within a shell with the environment variables present, but when they expire I have to restart the IDE to pick up the new token, which makes for quite an interruption in my dev workflow.

nickatsegment:
> I'd find this useful as well. My use case is running some integration tests from a Java IDE, where the integration tests require AWS privileges. I can invoke the IDE from within a shell with the environment variables present, but when they expire I have to restart the IDE to pick up the new token, which makes for quite an interruption in my dev workflow.

Can you not instruct the IDE to wrap the test command in `aws-okta exec`, rather than wrapping the whole IDE?

toddlipcon:
> Can you not instruct the IDE to wrap the test command in aws-okta exec, rather than wrapping the whole IDE?

Might be possible? I'm just using normal IDEA "run unit test" functionality here, and while I see ways to configure which JVM to use, I didn't want to go making custom JAVA_HOMEs with wrapper scripts in place of 'java'.

apechimp:
> So, I'll merge under the condition that @apechimp is the maintainer of this subcommand. Please add an entry for this file in CODEOWNERS.

According to the [code owners docs](https://help.github.com/en/articles/about-code-owners), code owners must have write access to the repository. I'm happy to maintain this command, but I want to verify that you actually want me to have write access to the repository.

> We need a better name that makes it clear its going to modify a file. I know I was the one who suggested `export`, but now it's actually updating a file. Maybe `update-credentials-file`? Kind of long-winded... open to ideas.

Seems fine. I'll make this change. In hindsight, `export` is kinda a bad name considering that `env` exports the credentials...

> Can you point ansible at a different path for `credentials`? Keeping temporary creds in a user-level config file feels a little weird to me. Not a huge deal.

I think so. The issue for me is that ansible isn't the only tool we use that uses our credentials file. We've for better or worse designed many of our aws workflows around switching profiles and expecting credentials to exist in the credentials file. 

> One more alternative: [boto3 (which I believe ansible uses) supports assuming roles itself, directly](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#assume-role-provider). Have you tried that?

So the issue here is that we use okta to sso into each account. We don't want to allow roles from one account to assume roles in another because we want to manage access to each account in one place: okta.

nickatsegment:
> According to the [code owners docs](https://help.github.com/en/articles/about-code-owners), code owners must have write access to the repository. I'm happy to maintain this command, but I want to verify that you actually want me to have write access to the repository.

Well dang, yeah, that's too much. Maybe just put a comment at the top of the file then.
 
> > Can you point ansible at a different path for `credentials`? Keeping temporary creds in a user-level config file feels a little weird to me. Not a huge deal.
> 
> I think so. The issue for me is that ansible isn't the only tool we use that uses our credentials file. We've for better or worse designed many of our aws workflows around switching profiles and expecting credentials to exist in the credentials file.

Gotcha. What do you think about making it a required argument? Say `aws-okta write-to-credentials <profile> <credentials-file>` e.g `aws-okta write-to-credentials profile ~/.aws/credentials`. That makes it pretty explicit that we're going to be modifying that file, and allows the security conscious to write it to an ephemeral file. Writing to a specific file might be used to avoid racing on the main file as well.

> > One more alternative: [boto3 (which I believe ansible uses) supports assuming roles itself, directly](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#assume-role-provider). Have you tried that?
> 
> So the issue here is that we use okta to sso into each account. We don't want to allow roles from one account to assume roles in another because we want to manage access to each account in one place: okta.

Understood, thanks for clarifying.

PR #201, opened 2019-08-27
feat: enable FIDO U2F MFA

a6sondhi:
adds support for U2F FIDO devices such as yubikey. The U2F device must
already be registered with Okta and be plugged in and available for use.

retry if open fails, there are issues with the low level `hid_open` call
that we rely on in `hidapi`.

These are the libraries we depend on:
https://github.com/marshallbrekka/go-u2fhost/blob/master/hid/wrapper.go
https://github.com/marshallbrekka/go.hid
https://github.com/signal11/hidapi

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -19,6 +19,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/segmentio/aws-okta/lib/mfa"
 	"github.com/segmentio/aws-okta/lib/saml"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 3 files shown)


a6sondhi:
Made the changes, feel free to merge whenever.

lib/okta.go
```diff
@@ -365,8 +368,34 @@ func (o *OktaClient) postChallenge(payload []byte, oktaFactorProvider string, ok
...
+			fidoClient, err := mfa.NewFidoClient(f.Embedded.Challenge.Nonce,
+				f.Profile.AppId,
+				f.Profile.Version,
+				f.Profile.CredentialId,
+				o.UserAuth.StateToken)
+			if err != nil {
+				return err
+			}
 
+			SignedAssertion, err := fidoClient.ChallengeU2f()
```
nickatsegment: `SignedAssertion` -> `signedAssertion`?

switj: good catch, updated.

PR #203, opened 2019-08-27
feat: adding current role as env var

a6sondhi:
Storing current AWS assumed-role as `AWS_ROLE`. This allows the user to
easily check the assumed role when running aws-okta.


nickatsegment:
Hm, looks like a bad merge https://github.com/segmentio/aws-okta/issues/206 Can you please resubmit? Make sure you're based off the latest master.

PR #204, opened 2019-08-28
feat: support for aws govcloud profiles

simran-ssk:
Currently this project assumes we would always want to use us-gov-west-1 as it is statically set https://github.com/segmentio/aws-okta/blob/master/lib/okta.go#L217. This PR adds support so it is region agnostic.


simran-ssk:
@nickatsegment I made those lmk if it looks good.

PR #210, opened 2019-09-05
feat: build with go 1.13

### LINKED ISSUE #209: Use go 1.13
https://golang.org/doc/go1.13

We support the same versions as go, so we should now support (and build-test against) 1.13 and 1.12 and drop 1.11.

> The new go build flag -trimpath removes all file system paths from the compiled executable, to improve build reproducibility.

Perhaps we can remove `xerrors` now? https://github.com/segmentio/aws-okta/blob/6658fa0f1149debe1f94a667115265733f474535/internal/sessioncache/store_singlekritem.go#L10

nickatsegment:
Needed to bump xerrors to fix their go1.13 support.

Should cut a release after. The xerrors bump should fix Rowno's problem with our aws-okta homebrew recipe failing to build.

Fixes #209


.circleci/config.yml
```diff
@@ -1,8 +1,25 @@
 version: 2
 jobs:
-  test:
+
+  test-build-golang-prev:
     docker:
       - image: circleci/golang:1.12
```
Rowno: Shouldn't you run the tests on `golang:1.13` too? 🤔

nickatsegment: I guess it couldn't hurt, but I haven't seen it be a problem yet, so we'll cross that bridge when we get to it :)

PR #213, opened 2019-09-17
fix: Persist device token cookie in keystore

Chippiewill:
When using multifactor authentication Okta expects the device token cookie `DT` in addition to the session id cookie `sid`. Without it Okta will force you to reauthenticate the MFA session (which aws-okta
responds to by reauthenticating entirely). This is frustrating as it means aws-okta effectively doesn't perform any caching of the Okta signin which means the user has to reauth anytime they want to refresh / change their aws role.

This change persists the device token the same way as the session ID.

Note: I'm neither a Go developer nor familiar with this codebase, I have literally just bodged this in so this could probably be done in a slightly tidier way than just extending the args / return values even further.

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -75,7 +75,7 @@ type OktaCreds struct {
 
 func (c *OktaCreds) Validate(mfaConfig MFAConfig) error {
 	// OktaClient assumes we're doing some AWS SAML calls, but Validate doesn't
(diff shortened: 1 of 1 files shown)


sdann:
@nickatsegment I've refactored @Chippiewill's patch so that no existing APIs change. PR #259 obsoletes this PR.

PR #215, opened 2019-09-23
fix: invalid request when assume_role_ttl >1h used from upstream source_profile

brandt:
## Background

When role chaining, AWS sets a [hard upper limit of 1h on the chained assume role TTL][1].  However, the upstream non-chained role can still have a TTL of up to 12h.

## Context

Prior to this PR, if the chained role did not have a value set for `assume_role_ttl`, aws-okta would look for it in the source profile. If the source profile set it to a value >1h, aws-okta would make a bad request resulting in:

```
ValidationError: The requested DurationSeconds exceeds the 1 hour session limit for roles assumed by role chaining.
	status code: 400, request id: 2e2a5e15-da98-4bd0-8457-6e95565b7fe2
```

## Change

Adds a `GetDirectValue` function to only search for the setting in the given profile (i.e. doesn't descend into `source_profile` or the okta profile). This is matches the behavior of [aws-cli][2] [which does not appear to perform a recursive search for missing values][3].

I've updated `updateDurationFromConfigProfile` to use this so that it only looks for `assume_role_ttl` in the top-level profile.

[1]: https://github.com/awsdocs/iam-user-guide/blob/8d78057/doc_source/id_roles_terms-and-concepts.md
[2]: https://github.com/aws/aws-cli
[3]: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -66,10 +66,14 @@ func sourceProfile(p string, from Profiles) string {
 	return p
 }
 
-func (p Profiles) GetValue(profile string, config_key string) (string, string, error) {
+func (p Profiles) GetValue(profile string, config_key string, recursive bool) (string, string, error) {
 	config_value, ok := p[profile][config_key]
 	if ok {
 		return config_value, profile, nil
+	}
+
+	if !recursive {
+		return "", "", fmt.Errorf("Could not find %s in %s", config_key, profile)
 	}
 
 	// Lookup from the `source_profile`, if it exists
--- a/lib/config.go
+++ b/lib/config.go
@@ -66,30 +66,37 @@ func sourceProfile(p string, from Profiles) string {
 	return p
 }
 
-func (p Profiles) GetValue(profile string, config_key string, recursive bool) (string, string, error) {
+// GetDirectValue looks for the given setting directly set in the specified
+// profile. Unlike GetValue, it does not descend into `search_profile` or
+// fallback to the `okta` profile.
+func (p Profiles) GetDirectValue(profile string, config_key string) (string, string, error) {
 	config_value, ok := p[profile][config_key]
 	if ok {
 		return config_value, profile, nil
 	}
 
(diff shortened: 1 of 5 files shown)


brandt:
@andy-v-h I think you accidentally stamped the upstream project, but thank you for the review all the same 😄

lib/config.go
```diff
@@ -66,12 +66,16 @@ func sourceProfile(p string, from Profiles) string {
 	return p
 }
 
-func (p Profiles) GetValue(profile string, config_key string) (string, string, error) {
+func (p Profiles) GetValue(profile string, config_key string, recursive bool) (string, string, error) {
```
nickatsegment: I think I'd prefer a new `GetValueRecursive` function 

https://github.com/segmentio/aws-okta/wiki/Style-Guide#api-compatibility

sdann:
@brandt We also have this issue and would like to see you PR get in. Will you be making the refactor changes requested?

lib/config.go
brandt: I started to make this change, but ended up creating a `GetDirectValue` function instead. Here's why:

I was calling the existing (before this PR) `GetValue` behavior recursive. Upon revisiting this, it occurred to me that it was a bit misleading because it doesn't descend past the immediate `source_profile`.

To address the underlying issue, I needed a function that only looked at settings set directly on the given profile, which is what `GetDirectValue` does. My most recent commit restores the original `GetValue` behavior, but I did make it use `GetDirectValue` under-the-hood to DRY things up a little. Happy to restore the old implementation if you'd prefer less indirection over DRY here.

brandt:
@sdann I just pushed a new commit that I think addresses the core ask. I'll try to keep an eye on this over the next few days in case other changes are needed.

nickatsegment:
https://github.com/segmentio/aws-okta/issues/278

PR #216, opened 2019-09-23
fix: don't force reauth when calling GetRoleARN

brandt:
The call to `p.GetRoleArn()` added in https://github.com/segmentio/aws-okta/commit/396d453 was forcing a separate auth event. That broke session caching in 0.23.0 and 0.23.1.

So the first run of `aws-okta` would auth twice, including a second 2FA prompt. Every subsequent call to `aws-okta` then also required 2FA confirmation.

This PR modifies GetRoleARN to take the creds as an argument.  Those creds are fetched immediately before anyways.

This also adds the `AWS_OKTA_ASSUMED_ROLE_ARN` and `AWS_OKTA_ASSUMED_ROLE` environment variables to the `env` subcommand so that it is consistent with `exec`.

This also addresses the issue reported in segmentio/aws-okta#213

### DIFF
--- a/lib/provider.go
+++ b/lib/provider.go
@@ -307,19 +307,18 @@ func (p *Provider) roleSessionName() string {
 	return fmt.Sprintf("%d", time.Now().UTC().UnixNano())
 }
 
-func (p *Provider) GetRoleARN() (string, error) {
-	creds, err := p.getSamlSessionCreds()
-	if err != nil {
-		return "", err
-	}
+// GetRoleARN makes a call to AWS to get-caller-identity and returns the
+// assumed role's name and ARN.
+func (p *Provider) GetRoleARN(c *credentials.Value) (string, error) {
 	client := sts.New(aws_session.New(&aws.Config{Credentials: credentials.NewStaticCredentials(
-		*creds.AccessKeyId,
-		*creds.SecretAccessKey,
-		*creds.SessionToken,
+		c.AccessKeyID,
+		c.SecretAccessKey,
+		c.SessionToken,
 	)}))
 
 	indentity, err := client.GetCallerIdentity(&sts.GetCallerIdentityInput{})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


nickatsegment:
This is awesome @brandt. Would you mind splitting the bugfix and adding the env thing in a separate PR please?

lib/provider.go
```diff
@@ -307,19 +307,18 @@ func (p *Provider) roleSessionName() string {
...
 }
 
-func (p *Provider) GetRoleARN() (string, error) {
-	creds, err := p.getSamlSessionCreds()
-	if err != nil {
-		return "", err
-	}
+// GetRoleARN makes a call to AWS to get-caller-identity and returns the
+// assumed role's name and ARN.
+func (p *Provider) GetRoleARN(c *credentials.Value) (string, error) {
```
nickatsegment: We should avoid breaking API, but considering this method is currently broken, I don't mind.

Also, since this method doesn't use `p`, maybe let's just make it a plain function.

nickatsegment:
Ah, I can kind of see how the env.go change is also a bugfix. I'm going to go ahead and add to your PR

nickatsegment:
Couldn't push to your branch for whatever reason so I just opened a new PR https://github.com/segmentio/aws-okta/pull/217

Chippiewill:
@brandt I'm not sure it necessarily addresses the same issue that my change (#213) fixes. This change appears to fix reuse of the root AWS session whereas my change fixes reuse of the Okta session.

This is an important distinction because in my environment the max session length for AWS is 1 hour whereas for Okta it's 10 hours.

brandt:
@Chippiewill I agree – I'll submit a new PR with different wording in the commit message so that GitHub doesn't [automatically close your PR](https://help.github.com/en/articles/closing-issues-using-keywords).

@nickatsegment I'll incorporate your changes from #217 as well.

brandt:
Closing in favor of #218

PR #218, opened 2019-09-24
fix: don't force reauth when calling GetRoleARN

brandt:
The call to `p.GetRoleArn()` added in https://github.com/segmentio/aws-okta/commit/396d453 was forcing a separate auth event. That broke session caching in 0.23.0 and 0.23.1.

So the first run of `aws-okta` would auth twice, including a second 2FA prompt. Every subsequent call to `aws-okta` then also required 2FA confirmation.

This PR modifies GetRoleARN to take the creds as an argument.  Those creds are fetched immediately before anyways.

This also adds the `AWS_OKTA_ASSUMED_ROLE_ARN` and `AWS_OKTA_ASSUMED_ROLE` environment variables to the `env` subcommand so that it is consistent with `exec`.

This replaces a couple other PRs:

Closes: #216
Closes: #217

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/99designs/keyring"
 	"github.com/alessio/shellescape"
+	"github.com/aws/aws-sdk-go/aws/credentials"
 	analytics "github.com/segmentio/analytics-go"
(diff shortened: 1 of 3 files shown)


nickatsegment:
Released in https://github.com/segmentio/aws-okta/releases/tag/v0.23.2

PR #219, opened 2019-09-24
feat: allow override of role arn to assume

switj:
This lets the user specify a role ARN that will override what is present
in their profile. If no value is present in the profile it will be
added.

Co-authored-by: Aaditya Sondhi <aaditya@autonomic.ai>

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -21,6 +21,7 @@ import (
 var (
 	sessionTTL    time.Duration
 	assumeRoleTTL time.Duration
+	assumeRoleARN string
 )
 
 func mustListProfiles() lib.Profiles {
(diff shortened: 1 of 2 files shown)


switj:
This PR is a followup to https://github.com/segmentio/aws-okta/pull/202 I've taken over the work from Aaditya who has returned to school.

switj:
> What's the use case here? Why not just write a new profile?
The usecase is to use role ARN with a generic okta profile (that doesn't provide a role ARN). This reduces setup requirements and consistency issues between different profiles. This feature would be predominantly used in scripts.

Our main motivation is to remove the prerequisite when running a script to have a specific profile setup. The only profile required is the `okta` profile that users setup when they install aws-okta.

Does this seem like a reasonable usecase to motivate adding this feature?

nickatsegment:
Actually yes, that sounds pretty nice actually. We distribute a set of standard profiles here but can never really guarantee they're up to date. And our tooling often references them. So yeah, it might be nice to not have to assume it's up to date.

PR #208, opened 2019-08-29
feat: adding current role as env var

a6sondhi:
Rebased to latest master and resubmitting https://github.com/segmentio/aws-okta/pull/203, which caused https://github.com/segmentio/aws-okta/issues/206.

PR #220, opened 2019-09-24
fix: make GetRoleARN support GovCloud

raylu:
GovCloud support was added in #197 / b4e7839 and #204 / 9787f11.
Taking the role as an env var was added in #208 / 396d453, which added
`GetRoleARN` which ignored region and, thus, broke GovCloud.
Both of these were released as v0.23.0, so no version of aws-okta
supported GovCloud.

Separately, `GetRoleARN` was split into a `lib` function and a `Provider`
function in #218 / e13ae0f, which left behind a code duplication TODO
and an unused Provider function.

### DIFF
--- a/lib/provider.go
+++ b/lib/provider.go
@@ -312,25 +312,41 @@ func (p *Provider) roleSessionName() string {
 	return fmt.Sprintf("%d", time.Now().UTC().UnixNano())
 }
 
+// GetRoleARN uses temporary credentials to call AWS's get-caller-identity and
+// returns the assumed role's ARN
+func (p *Provider) GetRoleARNWithRegion(creds credentials.Value) (string, error) {
+	config := aws.Config{Credentials: credentials.NewStaticCredentials(
+		creds.AccessKeyID,
+		creds.SecretAccessKey,
+		creds.SessionToken,
+	)}
+	if region := p.profiles[sourceProfile(p.profile, p.profiles)]["region"]; region != "" {
+		config.WithRegion(region)
+	}
+	client := sts.New(aws_session.New(&config))
+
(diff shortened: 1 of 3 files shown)


lib/provider.go
```diff
@@ -312,24 +312,18 @@ func (p *Provider) roleSessionName() string {
...
 
-// GetRoleARN uses p to establish temporary credentials then calls
-// lib.GetRoleARN with them to get the role's ARN
-func (p *Provider) GetRoleARN() (string, error) {
-	creds, err := p.getSamlSessionCreds()
-	if err != nil {
-		return "", err
+// GetRoleARN uses temporary credentials to call AWS's get-caller-identity and
+// returns the assumed role's name and ARN.
+func (p *Provider) GetRoleARN(creds credentials.Value) (string, error) {
```
nickatsegment: `lib.GetRoleARN` was added so we could pass the credentials in without breaking API compatibility of `Provider#GetRoleARN`. I realize the API is gross and probably nobody imports it externally, but we can't be sure.

How about just adding a new function and method `GetRoleARNWithRegion`? I realize this contributes to API sprawl, but we can tackle that in v3.0

raylu: done (and restored the old `lib.GetRoleARN`)!

raylu:
could we get a 0.24.2 release to pick this up? https://github.com/segmentio/aws-okta/compare/v0.24.1...master

eculver:
I think a release is in order but I think this might warrant a 0.25 release given that it's adding new functionality. @nickatsegment you have any issues w/ cutting v0.25.0?

nickatsegment:
Yeah, kind of on the fence on that one, but meh, minor versions are cheap

eculver:
v0.25.0 is out and contains these changes.

Issue #91, opened 2018-10-25
Support for multiple Okta tenants at once


devonbleak:
Would love to see support for multiple Okta tenants.

nickatsegment:
I believe this is solved by https://github.com/segmentio/aws-okta/pull/92? Ping me if not and we can reopen

devonbleak:
@nickatsegment updated the description to be more accurate in what I'm looking for.  We're about to be living in 2 Okta worlds at the same time :-)

nickatsegment:
Ugh, that sounds unfun. Will reopen. Godspeed.

LukaszRacon:
Implemented in #221

PR #221, opened 2019-09-25
feat: Support for multiple Okta accounts (add keyring key configuration per profile)

LukaszRacon:
Add ability for user to specify keyring key to separate provider data (including MFA and domain) per aws profile. Similar to #91 request.

Setup accounts:
```ini
aws-okta add --keyringKey=okta-creds-account-a
aws-okta add --keyringKey=okta-creds-account-b
```

Define keyring key for each profile:
```ini
[profile account-a]
aws_saml_url = home/amazon_aws/woezQTbGWUaLSrYDvINU/214
role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
okta_keyring_key = okta-creds-account-a

[profile account-b]
aws_saml_url = home/amazon_aws/adbcQTbGaDAA4rYDvINU/123
role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
okta_keyring_key = okta-creds-account-b
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,8 +129,8 @@ The configuration above means that you can use multiple Okta Apps at the same ti
 #### Multiple Okta accounts
 setup accounts:
 ```ini
-aws-okta add --keyringKey=okta-creds-account-a
-aws-okta add --keyringKey=okta-creds-account-b
+aws-okta add --account=account-a
+aws-okta add --account=account-b
 ```
 
 define keyring key for each profile:
@@ -139,12 +139,12 @@ define keyring key for each profile:
 # This is a distinct Okta App
 aws_saml_url = home/amazon_aws/woezQTbGWUaLSrYDvINU/214
 role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
-okta_keyring_key = okta-creds-account-a
+okta_account_name = account-a
 
 [profile account-b]
 aws_saml_url = home/amazon_aws/woezQTbGaDAA4rYDvINU/123
 role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
-okta_keyring_key = okta-creds-account-b
+okta_account_name = account-b
 ```
 
 #### Configuring Okta session and AWS assume role TTLs
--- a/README.md
+++ b/README.md
@@ -126,6 +126,27 @@ role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
 
 The configuration above means that you can use multiple Okta Apps at the same time and switch between them easily.
 
+#### Multiple Okta accounts
+setup accounts:
+```ini
+aws-okta add --keyringKey=okta-creds-account-a
+aws-okta add --keyringKey=okta-creds-account-b
+```
+
+define keyring key for each profile:
+```ini
+[profile account-a]
+# This is a distinct Okta App
+aws_saml_url = home/amazon_aws/woezQTbGWUaLSrYDvINU/214
+role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
+okta_keyring_key = okta-creds-account-a
+
+[profile account-b]
+aws_saml_url = home/amazon_aws/woezQTbGaDAA4rYDvINU/123
+role_arn = arn:aws:iam::<account-id>:role/<okta-role-name>
+okta_keyring_key = okta-creds-account-b
+```
+
 #### Configuring Okta session and AWS assume role TTLs
 
 The default TTLs for both Okta sessions and AWS assumed roles is 1 hour.  This means that aws-okta will re-authenticate to Okta and AWS credentials will expire every hour.  In addition to specifying the Okta session and AWS assume role TTLs with the command-line flags, they can be set using the `AWS_SESSION_TTL` and `AWS_ASSUME_ROLE_TTL` environment variables respectively.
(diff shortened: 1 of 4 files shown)


nickatsegment:
This is a cool idea, especially useful for people who have to test against Okta Preview :)

I would argue that the name `keyring-key` requires a little too much knowledge of aws-okta's internals. How about just calling it something like `okta-account-name`, then the keyring key is `okta-creds-<okta-account-name>` (or just `okta-creds` if no account name)?

Also, do you think we need separate AWS session caches per okta account? I don't think so, since profiles should be unique across all accounts.

LukaszRacon:
Thanks for feedback. Profile names are unique so AWS sessions should be fine. I renamed `keyring-key` to `okta-account-name` for AWS profiles. I used `--account` parameter for `aws-okta add` command.

ps-jay:
I can't get this working in my environment.  I've just brew installed aws-okta v1.0.0.

The scenario is: the same Okta domain is used for two different usernames - a regular user, and an admin user.

Both users have access to the same AWS tile, but if you've logged into Okta with admin you'll have additional roles available to you.

I've done the aws-okta add command twice like so:
```
aws-okta add
# add regular user
INFO[0011] Added credentials for user jayp

aws-okta add --account=admin
# add admin user
INFO[0021] Added credentials for user adm_jayp
```

Then I've edited my `~/.aws/config` file so that it looks like this:
```
[okta]
aws_saml_url = home/amazon_aws/<something>/<###>

[profile admin-role]
role_arn = arn:aws:iam::<something>:role/Okta-Admin
okta_account_name = admin

[profile regular-role]
role_arn = arn:aws:iam::<something>:role/Okta-Regular
```

Now, using the regular role works just fine:
```
aws-okta --debug exec regular-role -- aws sts get-caller-identity
DEBU[0000] Parsing config file /Users/jayp/.aws/config
DEBU[0000] Using KrItemPerSessionStore
DEBU[0000] cache get `regular-role session (<something>)`: miss (read error): The specified item could not be found in the keyring
DEBU[0000] Using aws_saml_url from profile okta: home/amazon_aws/<something>/<###>
DEBU[0000] Using okta provider (okta-creds)
DEBU[0000] domain: <mycompany>.okta.com
DEBU[0002] Failed to reuse session token, starting flow from start
DEBU[0002] Step: 1
DEBU[0005] Step: 2
DEBU[0005] Step: 3
DEBU[0008] Got SAML role attribute: arn:aws:iam::<something>:saml-provider/OktaAD,arn:aws:iam::<something>:role/Okta-Regular
DEBU[0008] Step 4: Assume Role with SAML
DEBU[0009] pOktaSessionCookieKey: okta-session-cookie
DEBU[0009] Writing session for jayp to keyring
DEBU[0010] Using session RGFF, expires in 59m59.575771s
{
    "UserId": "<something>:jayp@reecenet.org",
    "Account": "<something>",
    "Arn": "arn:aws:sts::<something>:assumed-role/Okta-Regular/jayp@<mycompany.org>"
}
```

But, using the admin role does not:
aws-okta --debug exec admin-role -- aws sts get-caller-identity
DEBU[0000] Parsing config file /Users/jayp/.aws/config
DEBU[0000] Using KrItemPerSessionStore
DEBU[0000] cache get `admin-role session (<something>)`: miss (read error): The specified item could not be found in the keyring
DEBU[0000] Using aws_saml_url from profile okta: home/amazon_aws/<something>/<###>
DEBU[0000] Using okta provider (okta-creds-admin)
DEBU[0000] domain: <mycompany>.okta.com
DEBU[0002] Got SAML role attribute: arn:aws:iam::<something>:saml-provider/OktaAD,arn:aws:iam::<something>:role/Okta-Regular
getting creds via SAML: ARN isn't valid
```

Seems like it's related to reusing a session token? .. I notice that in the admin example, it doesn't say "Failed to reuse session token, starting flow from start" ...  does that mean it's using the regular account's session token?

Thanks for any pointers here @LukaszRacon !

**Comment by nickatsegment** [comment]
@ps-jay I don't use this feature myself, but I think you're right with the cached token think. https://github.com/segmentio/aws-okta/pull/213/files#diff-5057c2ca588b57c0ada94b258bba5624R654 looks like it isn't namespaced by account or anything

**Comment by LukaszRacon** [comment]
@ps-jay looks like caching token #213 broke multiple account support. 
Issue is with `"okta-device-token-cookie"` being hardcoded. Switching it to `p.OktaAccountName+"okta-device-token-cookie"` should fix it. The other option would be to externalize it in a similar way to `okta_session_cookie_key`. `"okta-session-cookie"` may also need to be adjusted to `p.OktaAccountName+"okta-session-cookie"`.

Try this setting first:
```
[okta]
aws_saml_url = home/amazon_aws/<something>/<###>

[profile admin-role]
role_arn = arn:aws:iam::<something>:role/Okta-Admin
okta_account_name = admin
okta_session_cookie_key = okta-session-cookie-admin

[profile regular-role]
role_arn = arn:aws:iam::<something>:role/Okta-Regular
okta_account_name = regular
okta_session_cookie_key = okta-session-cookie-regular
```

If above doesn't work you might have to switch back to a previous release 0.27 (till caching token issue gets resolved).

ps-jay:
Thanks for the advice. I'll try the workaround and report back.

ps-jay:
The workaround of adding the `okta_session_cookie_key` name to the profile worked, thanks!

PR #229, opened 2019-10-15
feat: sort roles listed in roles prompt

badie:
- Break it down by account ID to make it easier to find roles to select.
  Roles will be displayed similar to list below
   ```
    Account: 111111111111
      0 - arn:aws:iam::111111111111:role/my-role-1
      1 - arn:aws:iam::111111111111:role/my-role-2
      2 - arn:aws:iam::111111111111:role/my-role-3

    Account: 222222222222
      3 - arn:aws:iam::222222222222:role/other-role-1
      4 - arn:aws:iam::222222222222:role/other-role-2

    Select Role to Assume:
   ```

- Fix error when provided profile does not have a role_arn
  - User will be prompted to choose role from list of roles.

### DIFF
--- a/lib/utils.go
+++ b/lib/utils.go
@@ -5,6 +5,8 @@ import (
 	"encoding/xml"
 	"errors"
 	"fmt"
+	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


lib/utils.go
```diff
@@ -91,9 +97,23 @@ func GetRole(roleList saml.AssumableRoles, profileARN string) (saml.AssumableRol
...
+		return roleList[i].Role < roleList[j].Role
+	})
+
+	var previousAccountID, currentAccountID string
+
 	for i, arole := range roleList {
-		fmt.Printf("%d - %s\n", i, arole.Role)
+		currentAccountID = accountIDFromRoleARN(arole.Role)
+		if currentAccountID != previousAccountID {
+			fmt.Printf("\nAccount: %s\n", currentAccountID)
```
nickatsegment: This account header feels a little redundant given you print the whole role ARN each time anyway?

badie: You are right, it is redundant. I'll update it.

badie:
Hi @nickatsegment 

Thanks for the review! I've made changes to the code to print out account id as the header and role name instead of role arn.

Output is now of format below.

```
Account: 111111111111
   0 - my-role-1
   1 - my-role-2
   2 - my-role-3

Account: 222222222222
   3 - other-role-1
   4 - other-role-2
```

PR #237, opened 2019-10-28
fix: Include the profileARN for the session cache key

ebabani:
The profileARN can be overriden as part of the exec call.
This fixes a bug with the following commands:

- aws-okta exec profile --assume-role-arn role1 -- command
- aws-okta exec profile --assume-role-arn role2 -- command

The second command would be run using role1 since the session cache
thinks there's already a valid session. The cache lookup logic does not
take the assume role override into account when looking for a session.

### DIFF
--- a/internal/sessioncache/key_profilearn.go
+++ b/internal/sessioncache/key_profilearn.go
@@ -1,0 +1,33 @@
+package sessioncache
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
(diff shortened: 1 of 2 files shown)


ebabani:
Sure, that change sounds good.

After the change nothing will be using OrigKey. Is there value in keeping it around?

nickatsegment:
> After the change nothing will be using OrigKey. Is there value in keeping it around?

You're right, probably not. But we can tackle that later.

PR #238, opened 2019-10-28
Helpful error when okta credentials not in keyring

pelletier:


lib/okta.go
```diff
@@ -598,6 +598,9 @@ type OktaProvider struct {
 func (p *OktaProvider) Retrieve() (sts.Credentials, string, error) {
 	log.Debugf("Using okta provider (%s)", p.OktaAccountName)
 	item, err := p.Keyring.Get(p.OktaAccountName)
+	if err == keyring.ErrKeyNotFound {
```
Fauzyy: @pelletier on second thought, could we move this error handling into the `cmd` package one level up where `Retrieve` is called. This feels like something the CLI client should be responsible for.

If you like, you could just add it to one of the commands to start and have a TODO there.

pelletier: Put here because the same handling was performed a few lines below. Happy to move both if you think it’s better.

Fauzyy: I think we can keep the unmarshal error here for now, and move the `ErrKeyNotFound` outside. What do you think?

pelletier: Took a look this morning. `Provider.Retrieve` is used 5 separate times in `cmd`:

Screen Shot 2019-10-29 at 10 04 06 AM

Pushing this error handling would probably need a bit of factoring to avoid copy pasting, which seems outside the scope of this PR. Happy to go either way!

PR #242, opened 2019-10-31
Use regional endpoints for STS #225

### LINKED ISSUE #225: aws-okta doesn't work with ap-east-1 region
```
aws-okta exec x -- aws ec2 describe-instances --region ap-southeast-1
{
    "Reservations": []
}
```

```
aws-okta exec x -- aws ec2 describe-instances --region ap-east-1

An error occurred (AuthFailure) when calling the DescribeInstances operation: AWS was not able to validate the provided access credentials
exit status 255
```

mikn:
After digging around a bit, I realised that you do not provide the `Endpoint` parameter to the `aws.Config{}` struct for sts. Because newer regions requires v2 tokens (see: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html and here: https://docs.aws.amazon.com/cli/latest/reference/iam/set-security-token-service-preferences.html) enabled on the global endpoint (and this is not enabled per default), using regional endpoints is safer as they're guaranteed to work with the region you're trying to access.

This fixes #225

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -218,7 +218,8 @@ func (o *OktaClient) AuthenticateProfileWithRegion(profileARN string, duration t
 	if region != "" {
 		log.Debugf("Using region: %s\n", region)
 		conf := &aws.Config{
-			Region: aws.String(region),
+			Region:   aws.String(region),
+			Endpoint: aws.String(fmt.Sprintf("https://sts.%s.amazonaws.com", region)),
 		}
 		samlSess = session.Must(session.NewSession(conf))
 	} else {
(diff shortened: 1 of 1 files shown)


mikn:
Thanks for merging! Ironically when I was digging even deeper into this, I found that they've added another configuration flag for managing this same behaviour (ref: https://github.com/aws/aws-sdk-go/releases/tag/v1.25.18) - but you are on a very old version of aws-sdk-go, so I'll open a new PR with an updated version of the aws-sdk-go and this new setting in use as well.

Issue #236, opened 2019-10-25
ATTENTION DEVS: moratorium on breaking changes until 2020-01-01


nickatsegment:
https://github.com/segmentio/aws-okta/issues/234 entails a fairly involved refactor. I'm giving @switj and their team 2 weeks to get it done. During this time, to make the refactoring easier and to ensure it's comprehensive, we  should avoid making any big changes, especially to the API.

If we find we absolutely must, the change must be listed here so that it can be incorporated into #234.

Updated to 2020-01-01. This is taking longer than expected. I'm going to take over the refactor now though.

ebabani:
@nickatsegment Mind if we get this small bugfix in? https://github.com/segmentio/aws-okta/pull/237

It makes the `exec -r` option work properly when switching between multiple roles.

nickatsegment:
@ebabani Yep, I had a comment on that PR, but yes, sounds like it's worth making an exception for. Thanks for following procedure ✔️

nickatsegment:
Merged https://github.com/segmentio/aws-okta/pull/245

nickatsegment:
Merged https://github.com/segmentio/aws-okta/pull/244

nickatsegment:
Merged https://github.com/segmentio/aws-okta/pull/255

nickatsegment:
Merged https://github.com/segmentio/aws-okta/pull/260 and https://github.com/segmentio/aws-okta/pull/259

nickatsegment:
https://github.com/segmentio/aws-okta/issues/278

PR #243, opened 2019-11-01
refactor lib to support other Okta Providers

switj:
create parralel lib2 that has "clients" and "providers" packages.

clients:
 - contains the components to interact with Okta
 - will cache your Okta session for you if a keyring is passed in

providers:
 - contains the components needed by Okta Providers. Currently this is
only the AWS SAML provider but could contain other providers in the
future.
 - will cache the session that this provider creates

Other than seperating logic this commit also introduces tests for
clients and providers as a starting point for more comprehensive test
coverage.


switj:
hi @nickatsegment I still have some additional work to do on this but I'm hoping it's close and won't need too much change  for the first iteration, I also wanted to give you as much time to review as possible.

I've put `[WIP]` in the title but any feedback you can give on what is there is greatly appreciated. I'm expecting there will be another (hopefully minor) update Monday so that I'll be able to remove the `[WIP]`.

One change I'd like to make is to drop the `lib2` and move `client/provider` to the top level or to `lib/`. What do you think?

Makefile
```diff
@@ -8,12 +8,20 @@ VERSION := $(shell git describe --tags --always --dirty="-dev")
...
-	GO111MODULE=on go test -mod=vendor -v ./...
+	GO111MODULE=on go test -mod=vendor -covermode=count -coverprofile=coverage.out -v ./...
+	@echo
+	@echo INFO: to launch the coverage report: go tool cover -html=coverage.out
+##
+## More information about cover reports:
+## https://blog.golang.org/cover
 
 all: dist/aws-okta-$(VERSION)-darwin-amd64 dist/aws-okta-$(VERSION)-linux-amd64
+linux: dist/aws-okta-$(VERSION)-linux-amd64
```
nickatsegment: 👍

Makefile
```diff
@@ -8,12 +8,20 @@ VERSION := $(shell git describe --tags --always --dirty="-dev")
...
 test:
-	GO111MODULE=on go test -mod=vendor -v ./...
+	GO111MODULE=on go test -mod=vendor -covermode=count -coverprofile=coverage.out -v ./...
+	@echo
+	@echo INFO: to launch the coverage report: go tool cover -html=coverage.out
+##
+## More information about cover reports:
+## https://blog.golang.org/cover
 
 all: dist/aws-okta-$(VERSION)-darwin-amd64 dist/aws-okta-$(VERSION)-linux-amd64
```
nickatsegment: This could now depend on `linux` and `darwin` instead of the actual files

lib2/provider/awssamlprovider.go
```diff
@@ -0,0 +1,450 @@
...
+
+type OktaClient interface {
+	AuthenticateUser() error
+	GetSessionToken() string
+	SaveSessionCookie() error
+	Request(method string, path string, queryParams url.Values, data []byte, format string, followRedirects bool) (*http.Response, error)
+	GetURL(string) (*url.URL, error)
+}
+
+type AwsSamlProvider struct {
```
nickatsegment: Let's take this opportunity to rename this `AWSSAMLProvider`

nickatsegment: And anywhere else where we have initialisms. https://github.com/golang/go/wiki/CodeReviewComments#initialisms

cmd/add.go
```diff
@@ -106,19 +107,19 @@ func add(cmd *cobra.Command, args []string) error {
...
+		Domain:   oktaDomain,
 	}
 
 	// Profiles aren't parsed during `add`, but still want
 	// to centralize the MFA config logic
 	var dummyProfiles lib.Profiles
 	updateMfaConfig(cmd, dummyProfiles, "", &mfaConfig)
 
-	if err := creds.Validate(mfaConfig); err != nil {
+	oktaClient, err := client.NewOktaClient(creds, nil, mfaConfig)
```
nickatsegment: Don't think you're handling `err` here?

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+	OktaServerPreview = "oktapreview.com"
+	OktaServerDefault = OktaServerUs
+
+	// deprecated; use OktaServerUs
+	OktaServer = OktaServerUs
+
+	Timeout = time.Duration(60 * time.Second)
+)
+
+type OktaUser struct {
```
nickatsegment: This is just used for deserialization, right? Just internally to `client`? Consider unexporting

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+	Username string `json:"username"`
+	Password string `json:"password"`
+}
+type OktaClient struct {
+	creds        OktaCredential
+	UserAuth     *OktaUserAuthn
+	DuoClient    *lib.DuoClient
+	SessionToken string
+	Expiration   time.Time
+	CookieJar    http.CookieJar
```
nickatsegment: Looks like this doesn't get used outside; unexport.

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+// Will fetch your Okta username, password, and domain from your keyring secret
+// backend.
+//
+// Will get the default credentials stored under the `okta-creds` key.
+//
+// feat-request: add support for getting additional sets of credentials.
+// The interface for this functionality needs to be defined, it's
+// possible to then implement alternative credential backends
+// to flexibly support alternative implementations.
+func GetOktaCredentialFromKeyring(kr keyring.Keyring) (OktaCredential, error) {
```
nickatsegment: Can we factor keyring out of the client code and into a simple interface? In the same way I did `sessioncache` I suppose.

Just thinking if you're using `client` from server code, you probably aren't going to want to use `keyring`.

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+// Supported configuration options:
+//       TODO: expand on configuration options and add tests.
+//
+// -- proxy config: TBD
+// -- session caching: Passing in a keyring will enable support for caching.
+//      this will cache a valid okta session securely in the keyring. This
+//			session is only for access to the Okta APIs, any additional sessions
+//			(for example, aws STS credentials) will be cached by the provider that
+//      creates them.
+func NewOktaClient(creds OktaCredential, kr *keyring.Keyring, mfaConfig MFAConfig) (*OktaClient, error) {
```
nickatsegment: This could just be called `New`

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+
+	if creds.IsValid() {
+		log.Debug("Credentials are valid :", creds.Username, " @ ", creds.Domain)
+	} else {
+		return nil, errors.New("credentials aren't complete. To remedy this, re-add your credentials with `aws-okta add`")
```
nickatsegment: Good opportunity to use typed errors here. Sketch:

```
type ErrorInvalidCredentials struct {}

...
return nil, &ErrorInvalidCredentials{}
...
```
Then in the CLI code:
```
var errInvalidCreds *client.ErrorInvalidCredentials
if errors.As(err, &errInvalidCredentials) {
  return errors.New("Invalid credentials; try `aws-okta add`")
```
That message is clearly for the CLI user, not the client user, so should be in the CLI.

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+
+	client := http.Client{
+		Transport: transCfg,
+		Timeout:   Timeout,
+		Jar:       jar,
+	}
+
+	oktaClient := OktaClient{
+		creds: creds,
+		// remove this now that we're storing the client
```
nickatsegment: Do this :)

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+		MFAConfig: mfaConfig,
+		UserAuth:  &OktaUserAuthn{},
+		Keyring:   kr,
+		client:    client,
+	}
+
+	// this can fail if we don't have have a backend defined.
+	// failing to retrived a cached cookie shouldn't fail the entire
+	// operation.
+	err = oktaClient.retrieveSessionCookie()
```
nickatsegment: Oof, probably not desireable

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+	if err != nil {
+		// log the error to debug and continue
```
nickatsegment: Oof, doubt this is desireable

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+	mySessionResponse, err = o.Request("GET", "api/v1/sessions/me", url.Values{}, []byte{}, "json", false)
+	if err != nil {
+		return
+	}
+	defer mySessionResponse.Body.Close()
+
+	// https://developer.okta.com/docs/reference/api/sessions/#get-current-session
+	// "if the session is invalid, a 404 Not Found response will be returned."
+	// checking for ok status (200) is adequate to see if the session is still valid.
+	sessionValid = mySessionResponse.StatusCode == http.StatusOK
```
nickatsegment: I wonder if this should just return an error if the session is invalid, instead of also having this bool

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+	if o.UserAuth.SessionToken == "" {
+		log.Debug("Auth failed. Reason: Session token isn't present.")
+		return fmt.Errorf("authentication failed for %s", o.creds.Username)
+	}
+	return
+}
+
+// Public interface to get the Okta session token.
+func (o *OktaClient) GetSessionToken() string {
+	return o.UserAuth.SessionToken
```
nickatsegment: Should we unexport `SessionToken` then?

nickatsegment: Or maybe `UserAuth` entirely

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+	}
+
+	return nil, fmt.Errorf("Failed to select MFA device with Provider = \"%s\", FactorType = \"%s\"", o.MFAConfig.Provider, o.MFAConfig.FactorType)
+}
+
+// Will prompt the user to select one of the configured MFA devices if an MFA
+// configuration isn't provided.
+//
+// TODO: convert this to a passed io reader to facilitate simple testing.
+func (o *OktaClient) selectMFADevice() (*OktaUserAuthnFactor, error) {
```
nickatsegment: I'd abstract this into some sort of `Chooser` interface and have the CLI code pass it in. Separates the UI from the business logic

lib2/client/okta.go
```diff
@@ -0,0 +1,667 @@
...
+//
+// as an example this would include sending a request for an SMS code.
+func (o *OktaClient) preChallenge(oktaFactorId, oktaFactorType string) ([]byte, error) {
+	var mfaCode string
+	var err error
+
+	//Software and Hardware based OTP Tokens
+	if strings.Contains(oktaFactorType, "token") {
+		log.Debug("Token MFA")
+		mfaCode, err = lib.Prompt("Enter MFA Code", false)
```
nickatsegment: Similarly, we could take a `MFACodeSupplier` interface or something

lib2/client/struct.go
```diff
@@ -0,0 +1,57 @@
+package client
+
+type SessionRequest struct {
```
nickatsegment: It's like most of these structs could be unexported?

lib2/provider/awssamlprovider.go
```diff
@@ -0,0 +1,450 @@
...
+type SessionCacheInterface interface {
+	Get(sessioncache.Key) (*sessioncache.Session, error)
+	Put(sessioncache.Key, *sessioncache.Session) error
+}
+
+type OktaClient interface {
+	AuthenticateUser() error
+	GetSessionToken() string
+	SaveSessionCookie() error
+	Request(method string, path string, queryParams url.Values, data []byte, format string, followRedirects bool) (*http.Response, error)
```
nickatsegment: This method is super broad. A quick look says we only call it twice, with `getAwsSAML`, and only the query param changes.

lib2/provider/awssamlprovider.go
```diff
@@ -0,0 +1,450 @@
...
+	GetURL(string) (*url.URL, error)
+}
+
+type AwsSamlProvider struct {
+	credentials.Expiry
+	AwsSamlProviderOptions
+	oktaClient             OktaClient
+	keyring                keyring.Keyring
```
nickatsegment: Had a quick look, and I think it really just needs a sessioncache, not the whole `kr`

lib2/provider/awssamlprovider.go
```diff
@@ -0,0 +1,450 @@
...
+	Expires                time.Time
+	sessions               SessionCacheInterface
+	defaultRoleSessionName string
+}
+
+type AwsSamlProviderOptions struct {
+	SessionDuration    time.Duration
+	AssumeRoleDuration time.Duration
+	ExpiryWindow       time.Duration
+	Profiles           lib.Profiles
```
nickatsegment: `Profiles` needs more structure and methods. Looking up things in a map is super lame

lib2/provider/awssamlprovider.go
```diff
@@ -0,0 +1,450 @@
...
+		SessionToken:    *(creds.SessionToken),
+		ProviderName:    "okta",
+	}
+
+	return value, nil
+}
+
+// GetRoleARN uses temporary credentials to call AWS's get-caller-identity and
+// returns the assumed role's ARN
+func (p *AwsSamlProvider) GetRoleARNWithRegion(creds credentials.Value) (string, error) {
```
nickatsegment: Using these AWS structs feels very weird to me. IMO, we should make our own.

As an example, we take in a `credentials.Value` here, which has a `ProviderName` field. We don't use that field in this method.

Not a huge deal.

nickatsegment:
> One change I'd like to make is to drop the `lib2` and move `client/provider` to the top level or to `lib/`. What do you think?

Yep!

lib2/client/okta.go
switj: We store it in the http client now. I've removed it.

lib2/client/okta.go
switj: I wanted to have a separate `bool` so that we can determine if the request is failing to complete vs the session not being valid. Not sure how important that is but that is why I split it out.

lib2/provider/awssamlprovider.go
switj: makes sense, I've updated it. Some of the commands still have flags for the session cache type to use that are being ignored. 

Should I remove the flags as well?

lib2/provider/awssamlprovider.go
switj: I'm trying to keep the client generic enough so it will work for many use cases, including the implicit oidc use case that I need.

lib2/client/okta.go
switj: considered renaming it but thought that `NewOktaClient` was more explicit. I don't have strong feelings either way.

lib2/client/okta.go
switj: This helper function is meant as a compromise for getting the okta crendentials from the keyring. Maybe it should be moved to `cmd`?

for the session caching I would prefer something like `sessioncache`, maybe if there is a session cache that returns `[]byte` instead of a struct?

something like:
```
type SessionCacheInterface interface {
  Get(key string) ([]byte, error)
  Put(key string, data []byte, name string) (error)
}
```

lib2/client/okta.go
switj: agree, handling the error and make more explanation.

lib2/provider/awssamlprovider.go
switj: I this needs changes, could we tackle this as part of another issue?

lib2/provider/awssamlprovider.go
switj: I'd like to delete this method, it seems like more trouble than it's worth.

Any objection?

switj:
I've tried to address a bunch of your comments with focus on trying to keep the exports to a minimum.

If we can address a few things as followup PRs I think it would help quality overall and provide some time for consideration. This is the list I have:
- Update error handling strategy for the client/provider (this will include changes in cmd and lib)
- put more structure around profiles
- change prompts for MFA and role selection to use an interface that is passed in. 
- start using a session cache interface for Okta sessions

I'd also like to see the test coverage higher, I think this can be achieved by adding more testing for each new feature/change.

Apologies if I've missed any of your comments or misinterpreted anything, kindly tell me where I went wrong.

lib2/client/okta.go
nickatsegment: Yep, I wasn't necessarily proposing copying `SessionCacheInterface` verbatim, just copying the pattern.

Moving keyring functionality to `cmd` seems sensible. It could be a subpackage or not, depending on how big it is.

lib2/client/okta.go
nickatsegment: I'd say typed errors are a better way to deal with that (along with `errors.Is/As`), but I can see that having a `bool` might be simpler.

lib2/provider/awssamlprovider.go
nickatsegment: Well it depends I suppose. After this refactor, are we expecting this to be `v1` of the API or still `v0`? You're going to be the consumer, so if we stick with `v0` and the API changes in a minor release, would that be painful for you?

lib2/provider/awssamlprovider.go
nickatsegment: Yep, if the CLI doesn't call it or you can simplify it somehow, by all means, go for it.

nickatsegment:
> If we can address a few things as followup PRs I think it would help quality overall and provide some time for consideration

The trouble is that some of these future changes would require changing the public API in backwards-incompatible ways. This is not the end of the world, considering we're v0, but limits its utility as a library, which was kind of the whole point, right? ;)

I don't think there's any real pressure to get this done immediately from my end, other than the fact that the longer we take, the more drift accumulates in `master` which will require merging. I'm totally happy extending the time in https://github.com/segmentio/aws-okta/issues/236 for another 2 weeks at least, if you have time.

That said, some specific comments:

> Update error handling strategy for the client/provider (this will include changes in cmd and lib)

Yeah, this probably wouldn't break the API (i.e., things would still compile) since any error-branching is runtime anyway. So this can be done later. (Even if we're returning an `error` now, that error in the future could have more concrete types added later.)

> change prompts for MFA and role selection to use an interface that is passed in.

Aren't you going to need to do this to make it consumable as library anyway? You don't want prompts popping up in your server code

> I'd also like to see the test coverage higher

Honestly increasing the test coverage is not high priority for me. At the end of the day, if current flows through the CLI work as they do right now, that should be good enough.

nickatsegment:
It's probably too hard to do now, but a commit history would have made this easier to follow :) It's hard to see what's changed since the last time I reviewed if you keep squashing. 

Related: hit that `Resolve Conversation` button if you think you've fixed the thing.

lib2/client/okta.go
switj: I'm implemented the session cache for OktaClient. 

Keeping the helper function where it is seems fine for now, I've reconsidered and it doesn't seem like we need to move it.

switj:
I've put in another commit with the updates, I'll refrain from the ammend/force workflow.

Please see the latest commit for what I added today, I'll continue tomorrow with anything that is still outstanding.

switj:
the failing build is caused by the introduction of typed errors and the us of `errors.Is`. I will look at fixing this tomorrow.

This introduces four errors:
```
InvalidCredentialsError = errors.New("Okta credentials are not valid")
InvalidSessionError     = errors.New("Okta session is not valid")
UnexpectedResponseError = errors.New("Okta returned an unexpected response")
NotImplementedError     = errors.New("Not implemented")
```

lib/client/okta.go
```diff
@@ -41,9 +40,19 @@ type OktaClient struct {
...
 	client       http.Client
+	selector     MFAInputs
+}
+
+type MFAInputs interface {
+	ChooseFactor(factors []MFAConfig) (int, error)
+	CodeSupplier(factor MFAConfig) (string, error)
+}
+
+type SessionCache interface {
```
nickatsegment: Ah yes, much better. 

I think we could shift some of the logic from here into the session cache itself. For example, knowing/generating the `key` I think should be the responsibility of the session cache, not the caller.

The `getSessionCookieKeyring` could move there as well.

Also, `data` is always just the session cookie, right? Might as well name it that way

The interface would then be something like

```
type SessionCache struct {
	Get(username string, domain string) ([]byte, error)
	Put(username string, domain string, cookie []byte) error
}
```

I wonder if `Get` and `Put` will ever need more info than the username and domain? If that's the case, you might want to pass in a struct that we can expand, to avoid having to change this API.

While writing this, I realized I'm basically just describing the other `sessioncache` :P

lib/client/okta.go
```diff
@@ -278,43 +285,33 @@ func (o *OktaClient) GetSessionToken() string {
...
-
-	factor, err := selectMFADeviceFromConfig(o)
-	if err != nil {
-		return nil, err
-	}
-
-	if factor != nil {
-		return factor, nil
+	factorsI := make([]MFAConfig, len(factors))
+	for factorIndex, factor := range factors {
```
nickatsegment: I wonder if this logic belongs in `ChooseFactor`?

cmd/utils.go
```diff
@@ -0,0 +1,42 @@
+package cmd
+
+import (
+	"github.com/manifoldco/promptui"
```
nickatsegment: fancy!

nickatsegment:
This is all shaping up very nicely. :D

The failing build is because we build for both go1.13 and 1.12 (the currently supported versions). `errors.Is` was introduced in 1.13. We should probably just use https://godoc.org/golang.org/x/xerrors until 1.14 is released (and we can drop support for 1.12)

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ jobs:
 
   test-build-golang-prev:
     docker:
-      - image: circleci/golang:1.12
+      - image: circleci/golang:1.13
```
nickatsegment: Ah, this is meant to compile for the previous release of go, since people might still be installing with `go get`. See https://github.com/segmentio/aws-okta/pull/243#issuecomment-552015513 for a workaround to the problem you were seeing compiling

switj: thanks, I read the comment after pushing the change. I'll revert the comment and use xerrors instead.

nickatsegment:
@switj ping me when you're ready for another review pass :)

switj:
@nickatsegment Just pushed a WIP commit for the MFA refactor. Hoping to finish it up tomorrow but it's close, missing code cleanup and more tests.

Interested to hear your thoughts.

lib/client/okta.go
```diff
@@ -286,33 +290,53 @@ func (o *OktaClient) GetSessionToken() string {
...
+	var availableFactors []types.OktaUserAuthnFactor
+
 	factors := o.userAuth.Embedded.Factors
 
 	haveMFAConfig = o.creds.MFA.Provider != "" && o.creds.MFA.FactorType != ""
 
+	for _, factor := range factors {
+		supported := false
+		var device mfa.Device
+		for _, dev := range o.mfaDevices {
```
nickatsegment: I'm not sure about this `Supported` interface. Looks like you're just doing a `switch/case` with extra steps?

Having a look at the implementations, the inputs they use to make the decision are only the factor type and the provider, right? So no need to expose them to the whole `OktaUserAuthnFactor` struct

lib/client/mfa/fido.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func (d *FIDODevice) GetId() string {
+	return d.id
+}
+
+// Supported will check if the mfa config can be used by this device
+func (d *FIDODevice) Supported(factor types.OktaUserAuthnFactor) error {
+	if factor.FactorType == "u2f" && factor.Provider == "FIDO" {
+		d.id = factor.Id
```
nickatsegment: Is this the only `Supported` that mutates its receiver? That seems odd

lib/client/mfa/fido.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil
+	}
+	return fmt.Errorf("FIDODevice doesn't support %s %w", factor.FactorType, types.ErrNotSupported)
+}
+
+// Verify is called to get generate the payload that will be sent to Okta.
+//   We will call this twice, once to tell Okta to send the code then
```
nickatsegment: I'm not really clear on this flow. The first time through, `d.codeRequested` will be `false`, so it gets set to `true`, then we return a blank `PassCode`?

nickatsegment: I think the separation of concerns between an `mfa.Device` and its caller needs to be more clearly specified. The fact that you're passing on-the-wire structs between them to me indicates that the Device is more information than it needs

lib/client/types/struct.go
```diff
@@ -1,70 +1,66 @@
-package client
+package types
```
nickatsegment: Hinted elsewhere, but I don't think this is a good idea. If you're sharing on-the-wire deserialization structs between packages, you're probably sharing too much information.

Besides, what if the JSON returned by Okta changes? Then you'll need to update the MFA code to deal with that, instead of the client

lib/client/types/errors.go
```diff
@@ -1,4 +1,4 @@
-package client
+package types
```
nickatsegment: Related to https://github.com/segmentio/aws-okta/pull/243/commits/5e41244927ee326b684b1d30d93adbbd342bfd52?file-filters%5B%5D=.go#r345545238, maybe these errors deserve their own package, but probably not.

nickatsegment:
> @nickatsegment Just pushed a WIP commit for the MFA refactor.

I like where your head's at. It would be nice if the MFA stuff didn't have to care about any Okta stuff, and be nice and self-contained. But I think your interfaces need a second pass (respectfully!). As written, I think they just add indirection rather than encapsulate complexity?

switj:
> > @nickatsegment Just pushed a WIP commit for the MFA refactor.
> 
> I like where your head's at. It would be nice if the MFA stuff didn't have to care about any Okta stuff, and be nice and self-contained. But I think your interfaces need a second pass (respectfully!). As written, I think they just add indirection rather than encapsulate complexity?

I appreciate your feedback, I'll review/considering everything carefully. From the first quick look I took it seems like you have some good points that need addressing.

This is the first pass, the plan is to take your feedback and get some additional feedback from others and make improvements to the interface today.

nickatsegment:
Hey @switj, thanks for taking the feedback graciously :) 

Considering all the hard work you've put into this, I was thinking I could send you some swag or something. My email address in on my profile if you're interested.

lib/client/mfa/fido.go
reidlevesque: It looks like we don't need this as we call `SetId()` from okta.go:314.

lib/provider/awssamlprovider.go
```diff
@@ -0,0 +1,472 @@
...
+		window = time.Minute * 5
+	}
+
+	// TODO(nick): why are we using the source profile name and not the actual profile's name?
+	source := lib.SourceProfile(p.profile, p.Profiles)
+	profileConf, ok := p.Profiles[p.profile]
+	if !ok {
+		return credentials.Value{}, fmt.Errorf("missing profile named %s", p.profile)
+	}
+	key := sessioncache.OrigKey{
```
ebabani: I believe this will get the wrong session since it doesn't take the role override into account. https://github.com/segmentio/aws-okta/pull/237

cmd/exec.go
```diff
@@ -239,7 +226,7 @@ func execRun(cmd *cobra.Command, args []string) error {
 
 	// Forward SIGINT, SIGTERM, SIGKILL to the child command
 	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, syscall.SIGTERM, os.Interrupt, os.Kill)
+	signal.Notify(sigChan, syscall.SIGTERM, os.Interrupt) // SA1016
```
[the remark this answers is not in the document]
switj: it's the code from staticcheck.

https://staticcheck.io/docs/checks

os.Kill cannot be trapped.

cmd/exec.go
```diff
@@ -145,7 +145,7 @@ func execRun(cmd *cobra.Command, args []string) error {
 	}
 
 	if _, ok := profiles[profile]; !ok {
-		return fmt.Errorf("Profile '%s' not found in your aws config. Use list command to see configured profiles.", profile)
+		return fmt.Errorf("profile '%s' not found in your aws config, use list command to see configured profiles", profile)
```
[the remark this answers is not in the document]
switj: I'm trying to follow golang best practices for errors, saying they shouldn't start with a capital.

more details here:
https://github.com/golang/go/wiki/Errors

lib/client/types/struct.go
switj: I understand where you're coming from but I think it is unavoidable if we want to have the interface be limited and allow MFA device implementations to be fully featured in the future. One example that passing the entire factor object allows is the ability for the sms MFA device to support re-requesting a code with MFA action `verify/resend`.

The MFA implementation are heavily reliant on the Okta API spec, if it changes then the MFA implementations will need to be updated. We're currently using the `v1` okta API so all device implementations must use that API.

lib/client/mfa/fido.go
switj: I've updated the logic to make it rely on the response from Okta rather than how many times Verify is called. Hopefully this is a better example of how we might use this Device interface and addresses your concerns.

lib/client/mfa/fido.go
switj: I've removed SetId/GetId and mutation in Supported for the FIDO implementation.

lib/client/okta.go
switj: It is essentially just a switch case but by using `Supported` we can allow external/custom implementations for `mfa.Device`. I've added two public methods to `OktaClient` to support this. 

`RegisterMFADeviceType` and `ClearMFADevices`

switj:
@nickatsegment ready for another review.

switj:
quick update, I now have a duo account and have it linked to our test okta account. I'm going to be testing the different factors to get this working again.

If anyone has experience using duo with okta and could help me it would be much appreciated. 
There is an official client that we might be able to use: https://github.com/duosecurity/duo_api_golang

nickatsegment:
@switj Sorry, I was on vacation last week. Will do another pass sometime this week while you work on the duo stuff.

switj:
@nickatsegment from what I can tell the duo mfa implementation is back to where it was before. I'm able to successfully auth using Duo `phone1` factor.

lib2/provider/awssamlprovider.go
switj: I think we're good to go ahead with a `v1` release.

lib/client/okta.go
switj: I've updated the logic for this a bit more, is the reason/value clearer now?

lib/client/types/errors.go
switj: If the errors types don't have their own package would we create separate errors for MFA and for client. I'm fine with either, was attempting to reduce duplication but I think the current split between client and client/mfa would do the trick.

What do you think?

switj:
@nickatsegment friendly reminder that this PR is still waiting for you to review. If I can do anything to make the review easier or if you're waiting on anything from me please tell me.

nickatsegment:
@switj Sorry, I've had to back-burner my OSS projects to focus. I'll have a look this week.

Makefile
```diff
@@ -8,12 +8,36 @@ VERSION := $(shell git describe --tags --always --dirty="-dev")
 LDFLAGS := -ldflags='-X "main.Version=$(VERSION)"'
 
 test:
-	GO111MODULE=on go test -mod=vendor -v ./...
-
-all: dist/aws-okta-$(VERSION)-darwin-amd64 dist/aws-okta-$(VERSION)-linux-amd64
+	GO111MODULE=on go test -mod=vendor -covermode=count -coverprofile=coverage.out -v ./...
```
nickatsegment: General comment: try to keep the PR focussed.  While this change is ok, it should have been in a separate PR. I know your intentions are good, but it makes it harder to accept the PR when there's orthogonal changes.

cmd/exec.go
```diff
@@ -239,7 +226,7 @@ func execRun(cmd *cobra.Command, args []string) error {
 
 	// Forward SIGINT, SIGTERM, SIGKILL to the child command
```
nickatsegment: Please update this comment to match https://github.com/segmentio/aws-okta/pull/243/files#r346352432

nickatsegment:
I started another review pass and realized it was gonna be a lot of persnickety little things that would drag this on even longer. The bones of this PR are great. Would it be ok if I just took your branch and did the changes on it myself? I'll start a new PR.

Just having a hard time convincing myself that this is going to be identical without writing some of it myself

switj:
go for it!

Makefile
switj: for these types of changes I can pull then out to another PR. I needed the changes while writing tests so I included it, it's not needed though.

nickatsegment:
Hey @switj, I spent some time trying to refactor to make it fit, but ultimately I came to the conclusion that the codebase we're working with is so riddled with technical debt and spaghetti code that a near total rewrite is in order.

I've started that here https://github.com/segmentio/aws-okta/issues/271 and would love any feedback you have along the way. Unfortunately, this rewrite may mean the lib API is fairly different.

I've already leveraged a bunch of your work here, and it was a huge help. So thanks for that.

nickatsegment:
Hey @switj, just wondering if you'd evaluated [saml2aws](https://github.com/Versent/saml2aws) as an alternative to aws-okta. I'm not sure how you're using us, but it looks like saml2aws is nicely structured and is consumable as a lib.

PR #245, opened 2019-11-04
fix: Produce prompts on stderr

asottile:
This enables scriptability of aws-okta, this appears to already be the case for some of the prompting things.

### before

(it hangs forever at the end of the output, I found it is prompting for the Role here)

```console
$ ./aws-okta exec "$ROLE" -- echo hi > log
INFO[0001] Select a MFA from the following list         
INFO[0001] 0: DUO (web)                                 
INFO[0001] 1: GOOGLE (token:software:totp)              
0
INFO[0004] Sending Push Notification...                 
INFO[0005] Device: phone1           
```

### after

(output is doctored a bit to obfuscate our account ids / etc.)

```console
$ ./aws-okta exec "$ROLE" -- echo hi > log
INFO[0001] Select a MFA from the following list         
INFO[0001] 0: DUO (web)                                 
INFO[0001] 1: GOOGLE (token:software:totp)              
Select MFA method: 0

INFO[0003] Sending Push Notification...                 
INFO[0004] Device: phone1                               

Account: 1234
   0 - bless-user

<< snip >>

Select Role to Assume: 0
$
```

### DIFF
--- a/lib/prompt.go
+++ b/lib/prompt.go
@@ -11,7 +11,7 @@ import (
 )
 
 func Prompt(prompt string, sensitive bool) (string, error) {
-	return PromptWithOutput(prompt, sensitive, os.Stdout)
+	return PromptWithOutput(prompt, sensitive, os.Stderr)
 }
 
 func PromptWithOutput(prompt string, sensitive bool, output *os.File) (string, error) {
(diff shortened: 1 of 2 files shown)


asottile:
thanks!

PR #250, opened 2019-11-21
fix: Update --assume-role-ttl default in README.md

douglaswth:


nickatsegment:
Fixed by https://github.com/segmentio/aws-okta/pull/255

PR #259, opened 2019-12-09
feat: Persist device token cookie in keystore to prevent repeated MFA prompt

sdann:
# Overview
 
Based on @Chippiewill PR #213
    
When using MFA, Okta expects the device token cookie `DT` in addition to
the session ID cookie `sid`. If MFA is required in the Okta app and this
cookie is not sent the session fails. This causes `aws-okta` to go
through the entire authentication flow again, eliminating any benefits
of session caching.
    
This change persists the device token the same way as the session ID.
    
I've kept the existing API identical, adding new function calls that
take an OktaCookies struct.

# Testing

Manual testing with Push notification and Okta AWS application setup to require MFA "Per Session" and "Once per Day".

With both patches in place, the Okta session is successfully reused, without MFA prompt, to create new AWS STS tokens when the previous STS token is expired.

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -495,7 +495,7 @@ func (o *OktaClient) challengeMFA() (err error) {
 
 	payload, err = o.preChallenge(oktaFactorId, oktaFactorType)
 
-	err = o.Get("POST", "api/v1/authn/factors/"+oktaFactorId+"/verify",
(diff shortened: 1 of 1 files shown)


sdann:
@nickatsegment any feedback on this PR?

PR #268, opened 2019-12-30
Cut v1.0.0

nickatsegment:
Following https://github.com/segmentio/aws-okta/wiki/v2.0.0-design-doc#rollout, this will be v1.0.0. v1.x will be maintenance only while we get v2.0.0 released.

I decided to abandon checking `vendor/` in. With the go module proxy added in 1.13 (and backported to 1.12), we should be safe against e.g. Github going down.

Everything is `internal` now; nobody should be able to import `aws-okta`. Kinda wish this was the default with Go (or there was a way to mark a module as non-importable entirely)


nickatsegment:
Reworked to get https://github.com/segmentio/aws-okta/pull/277 in. Turns out we do need vendoring for all the USB HID stuff, so I added it back.

Issue #248, opened 2019-11-13
FYI: Okta breaking change u2f to webauthn


marshallbrekka:
Figured I would file this as just an impending FYI.

Okta just transitioned our org from U2F factors to FIDO2 (WebAuthN).
This means that all our code that was looking for `u2f` factors suddenly broke.

I'm not sure if that change was specific to our org, or Okta wide.

For reference this is our internal change that does the bare minimum to mitigate the issue. It probably does not properly implement true `webauthn`, but you can use it as a guide for a U2F to WebAuthN migration path.

https://github.com/wearefair/okta-auth/pull/8/files

nijave:
We're seeing this issue as well

nickatsegment:
Oof, that's rough.

We don't use u2f/webauthn Okta internally at Segment (trust me, I'd like to), so we're relying on the community for a fix here.

dhess:
Yes, I noticed that when I asked Okta to enable the WebAuthn early access feature on my org's account, the U2F factor in MFA enrollment was replaced by a WebAuthn factor; i.e., it appears that you can enable either U2F or WebAuthn, but not both.

alsmola:
I've been working on supporting webauthn in https://github.com/Versent/saml2aws and I have an untested branch that could work for aws-okta:

https://github.com/alsmola/aws-okta/pull/1/files

I made some modifications to [`go-u2fhost`](https://github.com/marshallbrekka/go-u2fhost/pull/6) that I'm asking to be merged, but in the meantime this change uses my fork.

If someone can test and confirm this branch with `aws-okta`, I'm happy to hand it over.

PR #276, opened 2020-01-18
Migrate Factor U2F to WebAuthn

marshallbrekka:
Resolves https://github.com/segmentio/aws-okta/issues/248 in the 2.x RC.

```
$ ./aws-okta add -v marshallb fair.okta.com
INFO[0000] log level: info                              
Okta password: 
Validating credentials...
INFO[0003] requesting MFA                               

Touch the flashing U2F device to authenticate...
  ==> Touch accepted. Proceeding with authentication
Credentials validated!
Saved credentials to keyring for user marshallb@fair.okta.com (default)
```

PR #277, opened 2020-01-18
Bugfix to migrate Factor U2F to WebAuthn.

marshallbrekka:
Resolves https://github.com/segmentio/aws-okta/issues/248 against the 1.x code base.

https://github.com/segmentio/aws-okta/pull/276 Applies the same patch to the 2.x code.

I had to use https://github.com/goware/modvendor to vendor the required C files for the hid library, since `go mod` does not include them https://github.com/golang/go/issues/26366.

This is breaking the circleci verification of the mod vendor.
Would appreciate some suggestions for how the project would prefer to resolve that issue.


nickatsegment:
@marshallbrekka Thanks! Regarding the `go mod vendor` thing, I think we should make an exception for this. Just add some lines to the CI check to 1. install that tool and 2. grab the extra files.

marshallbrekka:
@nickatsegment updated

jspiro:
Has anyone confirmed this works? I created a test okta org with WebAuthn enabled, and an aws account with MFA enabled every sign on, and I cannot get it to work, fails on getting SAML creds.

I might have messed something up in configuring a test org – it's a ton of fiddly work. But I can confirm if I disable the MFA requirement then I am able to run my commands.

PR #280, opened 2020-02-15
fix: Bug 279: Clear the DT cookie when restarting AuthN workflow

sdann:
# Overview

In 1.0.0 the DT (device token) cookie was added, maintaining state on
the 2nd auth factor, between client and server. However, when the Okta
session expires, the old DT was still sent. The Okta server doesn't like
when an old DT, is used with a new session and re-prompts for 2FA with an
web form.

The aws-okta logic doesn't know about this 2FA web form, can't parse it,
and errors our.

We fix this bug, by clearing the DT cookie, whenever we restart the
AuthN workflow. Then the session and device token cookies are always in
sync between the client and server.

# Testing

Manually tested:
* aws-okta env - with out-of-date Okta session and DT
* aws-okta env - with garbage Okta session and out-of-date DT
* aws-okta env - with no stored Okta session or DT

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -229,6 +229,17 @@ func (o *OktaClient) AuthenticateProfile3(profileARN string, duration time.Durat
 	if err != nil {
 		log.Debug("Failed to reuse session token, starting flow from start")
 
+		// Clear DT cookie before starting AuthN flow again. Bug #279.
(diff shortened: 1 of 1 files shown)


nickatsegment:
https://github.com/segmentio/aws-okta/releases/tag/v1.0.1

Issue #119, opened 2019-02-05
Start providing Windows releases


devonbleak:
With the issues Go has on Windows with network drives and the recent addition of `gcc` as a build dependency and `libconv.dll` as a runtime dependency it's getting to the point that we can't reasonably expect users to all be building on Windows.

I'd be willing to start building/providing 64-bit WIndows binaries in a .zip or something if we can figure out how to incorporate it into the GitHub releases.

Ideally I think we'd get something like a .msi or chocolatey package built but that's probably beyond what I can commit to personally for now.

mlehner616:
+1

RafPe:
I would suggest then something really simple as https://github.com/goreleaser/goreleaser in conjunction with Travis CI. 

Have tested that and works flawlessly generating binaries for Linux/Windows and MacOS ( of course depending on config setup )

nickatsegment:
> I would suggest then something really simple as [goreleaser/goreleaser](https://github.com/goreleaser/goreleaser) in conjunction with Travis CI.

Currently we use CircleCI for CI, plain old go to do binary builds, and github-release to do the upload. I don't see an advantage to switching to goreleaser, though I haven't used it before. 

We'd have to add a new windows target in the Makefile. Because we rely on some C libs (USB HID) so I'm pretty sure we'd need to actually build on Windows, which [CircleCI only just added early-access support for on June 21](https://discuss.circleci.com/t/windows-early-access-now-available-on-circleci/30977), unless we figured out a cross-compile toolchain.
 	
Basically, this is all a bit of work and figuring out for a platform that Segment doesn't really use, and would be hard to offer any support or testing around. Hence the `help wanted` label.

dkujawski:
Hello, 

Are you still using CircleCI to manage your build processes?

I have been working on getting a CircleCI build setup to generate a Windows executable and MSI.  I have been able to successfully build the raw aws-okta executable that runs without the need to use the Win-builds process to install dependent libraries.  

I am currently working in a fork but would like to put a PR together once I get things cleaned up and the packaging, MSI, component in order.  

Aside from what is outlined in the contributing guidelines, is there anything that I need to be aware of or discussions that should be had?

-Dave

eculver:
Hi @dkujawski, thanks for helping out. I'll quote @nickatsegment from above
[...]

Issue #278, opened 2020-01-20
ATTENTION: aws-okta is on indefinite hiatus 


nickatsegment:
Hi all. Unfortunately Segment (and specifically I) will no longer be maintaining aws-okta, for the foreseeable future.

I tried to revive aws-okta with v2, but unfortunately it's just proven to be too much of a time sink to justify.

Several other OSS projects exist that can do what aws-okta does. [saml2aws](https://github.com/Versent/saml2aws) stands out in particular, with a nicer TUI, more providers, lots of MFA options, etc. I haven't done a security audit or anything, but at first blush it looks pretty competent and well-organized. It stores your session creds on disk instead of in a keyring, but I believe chaining this with [aws-vault](https://github.com/99designs/aws-vault) somehow would give you roughly the same level of protection.

I might also recommend forking aws-okta, perhaps starting from [my 2.0.0-rc1 branch](https://github.com/segmentio/aws-okta/pull/270). I couldn't detangle some of the lesser used features.

I'm going to leave Issues and PRs open for a while, but only for critical, straight-forward bug fixes. *No new features will be developed or added.*

[v1.0.0](https://github.com/segmentio/aws-okta/releases/tag/v1.0.0) is the final major release.

marshallbrekka:
Shameless plug:

Since I know some of the motivation for 2.x was to provide a pkg interface as well as a CLI, figured I would call out the project [okta-auth](https://godoc.org/github.com/wearefair/okta-auth). It provides an interface for driving the user authentication portion of the flow, resulting in an Okta session token.

It doesn't cover 100% of the functionality of aws-okta (notably obtaining AWS credentials from the Okta session), but can be used as a building block towards that goal.

It's used internally at [Fair](https://github.com/wearefair) to power our CLI for obtaining AWS credentials, as well as some internal application credentials not associated with AWS.

nickatsegment:
That's awesome; plugs encouraged! Part of the reason I had to give up was the feeling that with some work we could reuse parts of other FOSS instead of writing our own. But then `aws-okta` would just be a wrapper around the `pkg`s of `saml2aws` or something similar.

Personally it feels to me like everybody should probably just build their own in-house `aws-okta`-like. Making a tool to serve every use case is way too hard.

jspiro:
Who owns the brew package? Is there a chance segment will ever continue this, or will someone at segment be able to promote brew to a new maintainer when a fork ultimately takes over?

[...]

PR #281, opened 2020-02-19
Circleci windows build config

dkujawski:
In response to https://github.com/segmentio/aws-okta/issues/119

I was working on this before you guys shut things down and lost track of things in the shuffle.  I was doing some build related work in this area and picked it back up.  I lost my CircleCI access so unable to fully test recent changes.  But this config was working when I tested it a month or so ago. 

You don't need to build on a Windows machine, this works building in the same Linux environment that the the Linux build runs.  The previous issues with dependencies not being included seems to have been related to how the Go vendor logic is built to be shallow.  See: https://github.com/nomad-software/vend

The build process uses the `wixl` command from MSITools to generate the MSI file using the spec defined in the `WindowsMSI.wsx` file in the repo root.  See: https://wiki.gnome.org/msitools

It outputs a raw executable file along with an MSI.  The MSI will install the exe file into the Windows [AppDataFolder](https://docs.microsoft.com/en-us/windows/win32/msi/appdatafolder) for the user, update the Path environment variable and set appropriate values for uninstall etc. 

I followed existing patters in the `.circleci/config.yml` file to include the Windows build artifacts in the `github-release` process.

The Windows MSI build process is sensitive to version strings and requires semantic versioning.  Instead of doing fancy regex and parsing to extract the information from the `git describe --tags --always --dirty="-dev"` output I just set it in a `version` text file that gets parsed in the Makefile logic.  There is opportunity here to leverage github actions to keep the `version` file contents up to date as tagging happens.  Currently the contents of this file are static and needs to be updated to follow the current build version.  

_Another reason for the `version` file to exist, not related to CircleCI, but AWS CodePipeline processes were choking due to limited git metadata being available for various reasons.  The `git describe` command was failing.  Need to put a bunch of workarounds in place to get that working._

Just dropping this here just in case you guys want to include it.  I am not waiting for it.  We have our own build process running in AWS for the Windows stuff.  Just being a "good citizen" and passing it your way since the conversation was open.  

Happy to answer any questions, 

-Dave


nickatsegment:
Hey @dkujawski, thanks for this. I'm sure some Windows folks will find it useful. As you know, we're shutting down. Looks like the [community is rallying behind aws-okta/aws-okta](https://github.com/segmentio/aws-okta/issues/278#issuecomment-583548063) so you might want to cross-post there.

PR #283, opened 2020-03-02
Fixes #282 - aws-okta does not include OS in UserAgent string for Duo.

### LINKED ISSUE #282: DUO requires enabling "OtherOS" when using aws-okta
Hoping this meets the requirement of a straightforward bugfix.  Default user-agent string for Golang is simply "Golang-http-client/1.1" which is fine for many purposes, but Duo expect to see a an OS and Arch in parenthesis following it.  I wrote a library which solves this problem and makes the PR Into other apps a trivial matter.  I've only made the change to the duo.go code as this was the only place our corporate duo instance was blocking use of aws-okta.
Duo themselves have confirmed that the OS match for Mac is correct, I've asked them to review this PR and the library as well.

PR to follow.

beiriannydd:
User-Agent: Go-http-client/1.1 (Linux x86_64) aws-okta/1.0
User-Agent: Go-http-client/1.1 (Macintosh; Intel Mac OS X 10.14.6) aws-okta/1.0
User-Agent: Go-http-client/1.1 (Windows NT 10.0; Win64; x64) aws-okta/1.0

These are designed to be matches for existing browser OS identification strings.
The library uses
* The go http client User-Agent
* uname -a, sw_vers, cmd ver depending on architecture
* GOARCH
to produce the string, as well as the supplied client identifier of aws-okta/1.0 (which should help distinguish if someone picks up aws-okta/2.0!)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ require (
 	github.com/99designs/keyring v1.0.0
 	github.com/alessio/shellescape v0.0.0-20190409004728-b115ca0f9053
 	github.com/aws/aws-sdk-go v1.25.25
+	github.com/beiriannydd/useragent v1.0.2
 	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/karalabe/hid v1.0.0 // indirect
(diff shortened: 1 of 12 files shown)


nickatsegment:
@beiriannydd Thanks for the PR, but unfortunately I'm going to have to make the judgement that this is not a bugfix. It looks good though. Maybe [aws-okta/aws-okta](https://github.com/aws-okta/aws-okta) would take it.

PR #285, opened 2020-03-18
Update missing conversion of OktaAccountName

xarses:
GetSAMLLoginURL was using hard-coded "okta-creds" instead of the lookup
for OktaAccountName, in cases where multiple accounts are used, this can
result in building an incorrect login URL

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -698,7 +698,7 @@ func (p *OktaProvider) Retrieve() (sts.Credentials, string, error) {
 }
 
 func (p *OktaProvider) GetSAMLLoginURL() (*url.URL, error) {
-	item, err := p.Keyring.Get("okta-creds")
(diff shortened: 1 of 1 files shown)


nickatsegment:
This will mean existing users will have to re`add` their creds though I believe? 

Just trying to figure out which behaviour is less broken :)

xarses:
@nickatsegment no I didn't have to add either credential. I was able to recompile the binary and continue with my accounts as is.

Some more clarity on the problem I found was explicitly with `aws-okta login other-account` where one account was "default" i.e. no explicit account while `other-account` is explicitly another account. this results in `login` building a SAMLLoginURL that points to the default instance. This for some reason I haven't bothered to untangle does not happen if its a AssumeRole IAM just if its the SAMLAssumeRole landing account.

Other commands worked as expected without this modification i.e. `aws sts get-caller-identity` works with and without this change

Issue #298, opened 2020-09-28
Intermittent SAML and 2FA Push Notification Timeouts from Okta


sdann:
Posting to see if any other users are having similar issues. Since 9/24/20 several of our users are getting HTTP timeouts waiting for responses from Okta. The behavior is inconsistent, but it takes 3 forms:

1) Initial SAML Authn never responds. No MFA notification.

   `getting creds via SAML: Failed to authenticate with okta. If your credentials have changed, use 'aws-okta add': &url.Error{Op:"Post", URL:"https://convoy.okta.com/api/v1/authn", Err:(*http.httpError)(0xc0002fc0a0)}`

2) MFA prompt, but Push notification never shows up on mobile device. Eventual error:

   `getting creds via SAML: Post "https://[company].okta.com/api/v1/authn/factors/[id]/verify?rememberDevice=true": context deadline exceeded (Client.Timeout exceeded while awaiting headers)`

3) MFA notification arrives on phone and I complete it but `aws-okta` doesn’t seem to register it and it fails with timeout.

   `getting creds via SAML: Failed authn verification for okta. Err: Post "https://[company].okta.com/api/v1/authn/factors/[id]/verify": context deadline exceeded (Client.Timeout exceeded while awaiting headers)`

Right now, my assumption is this is a network level issue on Okta's servers, but it's only affecting `aws-okta`. Okta GUI authentication works fine.

nickatsegment:
Hm, haven't heard anything from our users at Segment.

It might be the case that the undocumented API we're using has changed in a subtle way that we're not hitting, like https://github.com/segmentio/aws-okta/pull/293 but we're not hitting the same codepaths somehow. Your only recourse is to compare HTTP flows between the two and reverse engineer it :(

Could also be network gremlins, as you suggest

brandt:
Several folks here have been running into this problem intermittently.

#299 does appear to fix it.

sdann:
I can confirm that HTTP2 over TLS is present in all of our failure cases, while HTTP1 over TLS is negotiated in all of my success repos.

In version 1.0.4, the `aws-okta` HTTP client offers that it can speak h2 or h1 in the Client Hello message.

![Screen Shot 2020-09-29 at 8 53 16 PM](https://user-images.githubusercontent.com/9983647/94641334-dca33c00-0295-11eb-9b9c-f218f832a6b3.png)

Even though the Get implementation explicitly requests HTTP/1.1.

[...]

PR #299, opened 2020-09-29
Force TLS negotiation to only support HTTP/1.1 - Issue #298?

mvallaly-rally:
Seems OKTA has a broken implementation of HTTP2?
There seem to be some quirks with they way golang handles HTTP2 requests (specifically when receiving GOAWAY frames) golang/go#20979, which result in the HTTP2 connection not terminating.

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -2,6 +2,7 @@ package lib
 
 import (
 	"bytes"
+        "crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -578,10 +579,19 @@ func (o *OktaClient) Get(method string, path string, data []byte, recv interface
 		}
 	}
 
+       // FORCE TLS negotiation to only support HTTP/1.1
+       tlsCfg := &tls.Config{
+                NextProtos: []string{"h1"},
+       }
+
 	transCfg := &http.Transport{
 		Proxy:               http.ProxyFromEnvironment,
 		TLSHandshakeTimeout: Timeout,
+                DisableKeepAlives:   true,
+                MaxIdleConnsPerHost: -1,
+                TLSClientConfig: tlsCfg,
 	}
+
 	client = http.Client{
 		Transport: transCfg,
 		Timeout:   Timeout,


brandt:
This does indeed appear to fix #298

lib/okta.go
```diff
@@ -578,10 +579,19 @@ func (o *OktaClient) Get(method string, path string, data []byte, recv interface
...
+       // FORCE TLS negotiation to only support HTTP/1.1
+       tlsCfg := &tls.Config{
+                NextProtos: []string{"h1"},
+       }
+
 	transCfg := &http.Transport{
 		Proxy:               http.ProxyFromEnvironment,
 		TLSHandshakeTimeout: Timeout,
+                DisableKeepAlives:   true,
+                MaxIdleConnsPerHost: -1,
```
nickatsegment, line 591: are these two settings also necessary?

sdann:
This fixes #298 in our environment as well, however, I've reduced it down to a minimal patch that still works:

```
diff --git a/lib/okta.go b/lib/okta.go
index f17cf86..a5a420d 100644
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -2,6 +2,7 @@ package lib
 
 import (
        "bytes"
+       "crypto/tls"
        "encoding/json"
        "errors"
        "fmt"
@@ -578,9 +579,13 @@ func (o *OktaClient) Get(method string, path string, data []byte, recv interface
                }
        }
 
+       tlsCfg := &tls.Config{
+       }
+
        transCfg := &http.Transport{
                Proxy:               http.ProxyFromEnvironment,
                TLSHandshakeTimeout: Timeout,
+               TLSClientConfig: tlsCfg,
        }
        client = http.Client{
                Transport: transCfg,
```

I don't yet understand why though.

sdann:
> I don't yet understand why though.

Default is HTTP1, says the Go docs.

https://golang.org/pkg/net/http/

```
    // TLSClientConfig specifies the TLS configuration to use with
    // tls.Client.
    // If nil, the default configuration is used.
    // If non-nil, HTTP/2 support may not be enabled by default.
    TLSClientConfig *tls.Config
```

Still, better to be explicit.

lib/okta.go
sdann: These weren't required in our repro. `DisableKeepAlives` makes the client setup a new TLS connection for each request. Though, that's the behavior I'm already seeing without it explicitly set.

nickatsegment:
Ok, I'm going to modify this PR to match @sdann's change and merge. It'd be nice to understand this behaviour better though

nickatsegment:
Superseded by #300

PR #300, opened 2020-09-30
Calculate OktaClient Content-Length correctly

Chippiewill:
Previously the content length was being calculated based off of an empty
uninitialised byte array. With this change it's now calculated off of the
actual data array used as the body.

Setting the content length to 0 seemed to be causing an issue with
recent changes to Okta's infrastructure as noticed in #298.

### DIFF
--- a/lib/okta.go
+++ b/lib/okta.go
@@ -553,7 +553,6 @@ func GetFactorId(f *OktaUserAuthnFactor) (id string, err error) {
 
 func (o *OktaClient) Get(method string, path string, data []byte, recv interface{}, format string) (err error) {
 	var res *http.Response
-	var body []byte
 	var header http.Header
 	var client http.Client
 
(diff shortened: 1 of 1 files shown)


nickatsegment:
Does this fix #298 even without #299 ?

nickatsegment:
@sdann Can you verify that this fixes #298 independent of #299?

Chippiewill:
Yes it does, I can up with this fix before seeing #299

sdann:
Wow, such a simple mistake. @nickatsegment confirmed, just this patch, without #298 fixes the issue in our environment.
h2 is still negotiated by TLS.

I'm guessing the HTTP1 server implementation handles the 0 Content-Length, while the HTTP2 implementation is stricter, and won't read any more of the POST packet because we said it was 0.

sdann:
We may also want to be explicit on the NextProto, but I'd love to get just this fix into an official 1.0.5 version to unblock my users ASAP.

nickatsegment:
@sdann awesome. Will get it merged ASAP

nickatsegment (approved):

nickatsegment:
Had a few Segmenters mention that they required this fix after coming back from vacation today, and that they were using v1.0.4 successfully up until then. Strange.

PR #303, opened 2020-10-22
Fix cred process expiration

reegnz:
The Expiration field conveyed the wrong value. It's changed to be the
same value that's used in aws-okta exec.

### DIFF
--- a/cmd/cred-process.go
+++ b/cmd/cred-process.go
@@ -122,8 +122,7 @@ func credProcessRun(cmd *cobra.Command, args []string) error {
 		AccessKeyID:     creds.AccessKeyID,
 		SecretAccessKey: creds.SecretAccessKey,
 		SessionToken:    creds.SessionToken,
-		// reuse the provided session duration
-		Expiration: time.Now().Add(p.SessionDuration).Format(time.RFC3339),
+		Expiration:      p.GetExpiration().Format(time.RFC3339),
 	}
 
 	var output []byte


nickatsegment (approved):

nickatsegment:
@reegnz what are the practical side effects you've been seeing?

reegnz:
What I've been noticing is that the timestamp returned by the cred process was not correctly represent the expiration of the credentials. I assume this broke the SDK credential refresh mechanism that actually runs cred-process *before* the expiration timestamp is reached.
For some reason this change is still not perfect (the timestamp still does not match the one in the token saved into the keychain (I'm on MacOS), so I think this fix is not all there is to it.

reegnz:
How I built around the faulty cred-process expiration up until now is to build my own cred-process script with bash utilizing aws-okta exec, and using the environment variables with jq to construct the correct json response.

reegnz:
One of the remaining issues that still needs to be fixed: When the SDK recognizes that the Expiration is approaching, it tries to refresh the token. aws-okta however sees that the token has not expired yet, returning the old token that's about to be expired.
I think the best fix for that would be to adjust the caching to be compatible with the refresh-mechanism in the AWS SDK.

reegnz:
Did dig a bit in aws-okta code further, just to make sure I was right with the SDK expiration.

Turns out this fix will be good enough. The Expiration comes from the AssumeRole that is performed when there's a source_profile in the config, so the expiry is now handled by the SDK correctly.
The AssumeRole is performed on each new exec or cred-process call, so those tokens are not cached. This allows the SDK to properly handle the rotation.

My initial assumption for looking into this came from how aws-vault does it: there it caches the AssumeRole credentials as well, so it makes the SDK credential refresh more difficult.

Here we don't have that issue present. :)

nickatsegment:
Super, thanks for the digging. FWIW, given #278, there's a certain point where a fix becomes complex enough that I couldn't justify accepting it. It might be worth it for you to invest your time in the fork at https://github.com/aws-okta/aws-okta, though I'm not seeing a lot of movement there.

PR #309, opened 2021-01-04
Update keyring to v1.1.6

Chippiewill:
Recent versions of kwallet have removed the old support for the kde4
compatible kwallet dbus interface. This means newer kde5 based
OS installs (e.g. kubuntu 20.04) can no longer use the kwallet backend
with aws-okta.

This was fixed upstream in the keyring lib back in 2019 but the
dependency hasn't been bumped since then.


nickatsegment (approved):
Dope.

nickatsegment:
I'm gonna consider this a "ship it and see" merge. If it breaks anybody, we'll immediately revert.

